PR #15, opened 2018-04-17
WIP: Using Gatsby.js

iddan:
- Replaced Jekyll with Gatsby
 - Moved index.md to documents/
 - Using official TC39 logo
 - Changed theme colour to match TC39 logo
 - With Gatsby the site is a statically generated PWA if you're wondering why it uses JS at all:
> Why remove JavaScript? That disables your React components on the client, slows down page transitions, prevents service workers from preloading page data, etc.
>
> In 0.x there's an command line flag to prevent the JavaScript from being written but I'm not inclined to support that in v1 as with code splitting, the amount of JS being loaded is quite small.
>
> There's always the option though of just not putting the script tags in the HTML.

https://github.com/gatsbyjs/gatsby/issues/962#issuecomment-301392995

#### TBD
 - [x] Define deployment
```bash
npm run deploy
# or
yarn run deploy
```
 - [ ] Silence Autoprefixer grid error: it's mad because IE doesn't support `grid-row` well


gatsby-config.js
```diff
@@ -0,0 +1,22 @@
+module.exports = {
+  siteMetadata: {
+    title: 'TC39',
+    lang: 'en-US',
```
xtuc, line 4: That reminds me that we didn't talk about the website translations. Do we even need them?

package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "gatsby-starter-default",
+  "description": "Gatsby default starter",
+  "version": "1.0.0",
+  "author": "Kyle Mathews <mathews.kyle@gmail.com>",
```
xtuc: Just a reminder, could you please clean this a bit?

xtuc (commented):
Looks cool alredy

gatsby-config.js
littledan: Translations would be great! Although it would be impractical to interact with TC39 in other languages, a translated introduction for generalist audiences would be really helpful. We have a big US-centric bias.

package.json
iddan: Of course! I missed that

gatsby-config.js
iddan: There are great solutions for multi-lingual. I think it's important but maybe not relevant right now as we are deploying an MVP

src/layouts/index.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react'
```
codehag, line 1: ops. So -- since you came late you might have missed this context -- we want this to be a static site, this is why we are using a static site generator such as gatsby. We do not want react or any other JS framework in place. Can you please refactor this so that these are either html files or MD files? Thanks!

codehag (changes requested):
Thanks for jumping on this! One issue that I noticed is that you introduced react. We do not want to use javascript to render our html. We should rely on html for layouts and md for content files. This is what was being done with the jekyll site. 

i just realized that gatsby is forcing us to use react and im afraid this rules out this framework :(. It does generate static pages but I do not feel comfortable with this since it privileges one framework over others, with little to no gain over just writing HTML. I think this sends a bad signal. When you said that it was a js generator I expected that it would allow us to write any type of JS. lets explore other options.

This site is primarily an information site. At the moment, we have no use for javascript. Lets keep the MVP in mind!

codehag:
Another note here. You did lots of great work outside of gatsby. it would be great to have these as separate prs so that the issue with gatsby doesn't stop the other work from getting in! thanks a lot for jumping on this!

codehag:
Also, really sorry about not looking at gatsby in more detail before :( that was my bad

xtuc:
Also FYI waiting for my flight I made https://github.com/xtuc/link-to-babel-repl-generator

iddan:
React is only used for declaring components and composing them the final output is a static HTML site so I don't see any problem with this. I don't see why using one library or another is problematic. Jekyll is a ruby library - why is it okay to be used but React not?

iddan:
> Another note here. You did lots of great work outside of Gatsby. it would be great to have these as separate PRs so that the issue with Gatsby doesn't stop the other work from getting in! thanks a lot for jumping on this!

Actually except for porting everything to Gatsby and removing Jekyll specific files I only replaced the logo and theme colour so I don't think this requires a whole different PR.

src/layouts/index.js
iddan: This is a static site. React is only used as a syntax for declaring and composing components.

iddan:
If we're going to have interactivity in the site later Gatsby provides excellent rehydration mechanism but as for now the `<script />` tags can just be stripped out and you have a perfectly static site.
The advantages of using React + Gatsby + GraphQL is that resources are bound to view **explicitly* and common components can be shared easily (unlike in Jekyll for instance).
Also as I pointed out earlier deploying a static site that it's scripts making it a PWA **boosts** performance so I don't see why not using this mechanism if it benefits the users.

gatsby-node.js
```diff
@@ -0,0 +1,7 @@
+/**
```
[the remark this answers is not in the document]
iddan: This will be used later to generate static pages

codehag:
We discussed a static site generator that would use MD files and be shipped as an MVP. 

At no point did we discuss using react, jsx, node, or graphql. At no point did we discuss PWAs. And yet, this PR is making all of those decisions for us. With Jekyll, ruby is an implementation detail. The focus is on the site organization and the content. Not the technology. Introducing Gatsby is unnecessary and causes more issues than it solves. 

I spoke with a few framework authors today. While the content is more important than the technology decision, this technology decision would limit our contributors to React developers. The JavaScript ecosystem is more than only React developers, and I feel that it is problematic to make a site that doesn't need any JavaScript rely on react. 

I would also like to have clearer communication in the future. In our discussions you mentioned gatsby as replacement for Jekyll, and I expected it to be that, then I saw this. This is doing much more than Jekyll. With Gatsby, React becomes the default framework, rather than javascript becoming an implementation detail like ruby does for Jekyll. This was not communicated. Had I known this, I would have raised the issue earlier.

Again, we never discussed using react, jsx, graphql, or node. What we discussed was a simple static site generator as part of an MVP, the primary goal of which is to show text. That is not to say that we will not -- in the future -- consider these. There may very well be benefits. However we are not doing this now. I expect this to be something we discuss when that time comes, not something a person makes an executive decision about. I hope this is clear.

Please split out the other changes from this pr, because those would be good to have. As a best practice, we shouldn't mix different types of fixes in the same pr anyway.

zoepage:
I think this discussion gets a little bit to heated for my taste. This is why I'd like the both (@iddan and @codehag ) of you to step away from it for a bit as this is violating the code of conduct regarding constructive communication. 

Thank you.

iddan:
I'm confused, it was a healthy discussion IMO. I was just trying to poke around to understand the requirements and restrictions Gatsby used as a test case. But maybe it was too intense for others IDK.

codehag:
Interesting. it is quite light weight which is great, uses javascript, and open to other technologies which is a plus. It doesn't tie us down very much which i think is important at this stage. 

i am cautious about using something hand-rolled, but since this is an interim solution i think it should be ok for now. What do others think?

thanks for the work you put into this @iddan!

littledan:
I'm fine to defer to @iddan here about any of the three choices. This is a small site and it won't be the end of the world if we have to move to a different system later.

iddan:
Personally I prefer Gatsby than investing in a new system (though developing something new is always fun). Jekyll is a non-option IMO because:
1. It's not safe enough
2. It's hard to config and learn
3. It's written in Ruby 4COL

codehag:
adding normalize has been broken out into this pr: https://github.com/tc39/tc39-web-draft/pull/23

codehag:
color and logo have been broken out into a separate pr: https://github.com/tc39/tc39-web-draft/pull/24

codehag:
Since we are deferring any changes until later, and this is a bit out of date, I am closing this for now

PR #24, opened 2018-05-04
Update colors and add logo asset

codehag:


assets/css/style.scss
```diff
@@ -16,7 +16,7 @@ a {
 
 .tc39-box {
   border: transparent 1rem solid;
-  background-color: #ff9900;
+  background-color: #FA7C23;
```
xtuc: `#fc7c00` is the official color of the logo? (according to the svg)

codehag: thanks!

littledan: I don't understand this `.tc39-box` class and div. Why don't we just reference the SVG file instead?

codehag:
That is the plan! But the svg was added in a different pr so that will be
done outside of this one

PR #28, opened 2018-05-22
Rebased add proposal table

codehag:
making this pr so that this can be presented to the tc39


codehag:
merging, as we have @Regaddi working on the css and it would be better for them to have access to the content of this pr

Issue #5, opened 2018-03-12
Footer with links


iddan:
It's a common pattern to have a footer with all the main internal and external links of the site.

codehag:
on this note, i like what they did here: https://webcompat.com/

zoepage:
A footer is a great way to sum up the most important links in one spot or make specific ones to stand out. After going though the information architecture with @codehag, I am not sure if we would need it. 

The main idea right now is to group the site content in 4 links:
- TC39 (about the group itself)
- specs (how to spec etc.)
- contribution (ways to contribute)
- state of proposals (list of proposals and their state)

The site content would be under those four and if necessary supported by a dropdown sub-navigation in the main-navigation.

Which links would you see in the footer or feel like should stand out that would not fit the schema above? I definitely would add e.g. twitter / github icons in the footer.

codehag:
@zoepage thank you so much for adding details to this!

zoepage:
Closing as we probably will not need a footer right now. We can reopen when we need it.

zoepage:
@iddan would you please be able to give a little bit more context here?

iddan:
I don't understand why footer is not needed, this is a draft site but it should simulate a real site as much as possible in terms of layout

zoepage:
Form follows function. As mentioned above we have 4 main links that group external links. The amount of the links is not verify the need of a footer, which is +7. Just having a footer for the sake of it should not be what we go for.

We can always add one as we needed as I mentioned above. Right now we want to have a stable site (MVP) that we can release, so lowest amount of features possible.

codehag:
Ah, i thought i reviewed this on the weekend.

It looks great! thanks for all the work!

PR #32, opened 2018-05-25
Fixes #12 - Implementation of design proposal

### LINKED ISSUE #12: [proposal] Design update
This is a proposal to update the current design including changes in the information architecture.

Above the fold, we'd have the 3 main topics that include the 4 main navigation points.
1. TC39 (about the project)
2. specs / contributie
3. state of proposal

The 3-grid copy is a "show-stopper" that would be supported though illustrations or icons and should give the user a first idea of the content of the site. The copy also gives various entry points though links to the internal content.

tc39_black

The navigation on the bottom of the page indicates the lower emphasis as the user should be focusing on the content above fold. While you scroll below the fold, the navigation bar would scroll up until it reached the top. It would stick to the top then and change the color to orange background and dark typo(#222). It would and also stay like this on every sub page.

I'll make more mocks for tablet and mobile view in the next few days.
Please note, this is a mockup and there are still small issues that will be fixed later on.

Thanks!

Regaddi:
This should be a base, where we can start finetuning.

I implemented the basic design concept in a responsive manner.
It should work from ~300px browser width upwards. Maximum content width is 1280px.

This is the stuff, I've done besides implementing the design proposal #12:
- Some minor transitions on the menu and accordion
- accessible with keyboard navigation
- accessible without JavaScript
- added and configured prettier and stylelint for codeformatting

I'm not 100% sure about the semantics of the layout. I adjusted the outline of the headings, but the page is now missing a `<h1>`.

I haven't touched the footer area, because of #5.

What do you think? @codehag @zoepage


codehag:
This looks great! There are three items missing, which can be added back in

Above the fold:
(here is the mock up https://user-images.githubusercontent.com/562354/40179919-c2472e26-59e5-11e8-9660-13b530d3d5ed.png)
- [ ] the header "specifying JavaScript" -- in the designs it says ECMAScript, but this should be JavaScript ;)
- [ ] the short form content, which is in three boxes above the fold. It has the same text for now as what you have for the long form content 

- [ ] Links out of the page that appear as drop downs from the menu. you should be able to see the few links that we have in the old version of the site! 

I am not too sure what should happen in the mobile version when we have everything lined up, @zoepage do you have any thoughts about this?

codehag:
Looks great, thanks for the work!

Issue #35, opened 2018-05-29
Improve readme of project


codehag:
Since we are going to make this available publicly, our readme could use a little work:

- [ ] double check the docker configuration
- [x] add a description of the site and its purpose
- [x] include a link to the code of conduct
- [x] write up a CONTRIBUTING.md doc (atom has a nice example: https://github.com/atom/atom/blob/master/CONTRIBUTING.md)
- [ ] add a license

PR #37, opened 2018-05-29
Create LICENSE

xtuc:
#35

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2018 TC39 members
+Copyright (c) 2018 ECMA's TC39 Royalty Free Task Group
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2018 TC39 members
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2018 TC39 members
```
xtuc: @codehag who should have the licence for that?

codehag: Not sure, @ljharb @littledan what do you think?

ljharb: that's a good question; generally the license needs to come from the contributors, so really i think it'd just be "contributors", and then whoever shows up in the commit log would be included?

Alternatively, it could probably be "ecma and TC39 members". It might be good to get a legal perspective.

codehag: @xtuc would you be able to make this update sometime in the next day? would be great to have it!

xtuc: What about "TC39 members and other contributors"? I doesn't really involves ecma?

codehag: sounds good

xtuc: I'm confused, what should I write?

rkirsling: Isn't "TC39 members" a bit odd if Ecma has 'members' and TC39 has 'delegates' though?

FWIW, it looks like test262 has a [BSD license](https://github.com/tc39/test262/blob/master/LICENSE) to `Ecma International` and an [open discussion](https://github.com/tc39/test262/pull/851) in which a concern about it has been escalated to Ecma. Perhaps this should be escalated too?

leobalter: For some formality, how about "Ecma's TC39 Royalty Free Task Group"? 

https://www.ecma-international.org/memento/TC39-RF-TG.htm

It can't go wrong this way.

codehag:
thanks!

littledan:
How about we use Ecma's software license, as test262 does? http://www.ecma-international.org/memento/TC39%20experimental%20policy.htm

codehag:
That link redirects rather quickly to another page. I am not too sure where to find the text you are referring to...

littledan:
Yeah, this should be improved... I will contact the Ecma Secretariat for next steps.

littledan:
Here's a better link--my mistake! https://www.ecma-international.org/memento/Policies/Ecma_Policy_on_Submission_Inclusion_and_Licensing_of_Software.htm

PR #52, opened 2018-06-03
make external links open in a new tab

codehag:


ljharb:
Is this really necessary? I find it very user-hostile when links open in a new tab unbidden, and every browser has “open in new tab” functionality for when i want this behavior. With the current behavior, i can also click “back” if i accidentally click away from a page i want to stay on.

Regaddi:
Besides the input by @ljharb: If using `target="_blank"` there should also be `rel="noopener noreferer"`. 😸 

See https://dev.to/ben/the-targetblank-vulnerability-by-example

codehag:
thanks for the feedback!

I was thinking about this, and left it here for feedback because I find it surprising if a website I am looking at redirects to another website that isn't part of what I was just looking at. but that is probably just me!

PR #54, opened 2018-06-04
Fixes #53 - Automatically run npm install on make install

### LINKED ISSUE #53: Automatically setup npm dependencies
We should automatically install the npm dependencies, when someone executes `make install`.
I'm not sure how to handle the case, when npm is not installed on the users machine or not the right version.
But this would automatically activate husky, lint-staged & stylelint, so we don't have to deal with code formatting in PRs.

Ideas? @codehag

Regaddi:
This will enable linting on a contributor's environment from the very beginning.

@codehag

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,9 @@ install: docker-build
 	docker start tc39-website
 	docker exec -ti tc39-website bundle install
 	docker stop tc39-website
+	@if command -v npm; then\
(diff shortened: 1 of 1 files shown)


ljharb:
The change is fine, but the question in #53 remains: why is `npm install` not sufficient by itself?

PR #65, opened 2018-06-13
fix: always use absolute URLs for assets, including protocols

### LINKED ISSUE #64: Icons not visible
When accessing the website with a bad network connection the icons are potentially invisible. The user might miss that there is a navigation icon, because the icon itself could not load.

We should provide a fallback by using native symbols.

Proposals:

Menu icon: ☰ (`&#9776;`)
Accordion caret: ▼(`&#9660;`)

Edit: they seem to be missing completely right now 🤔

Edit2:
missing on tc39.github.io/beta
visible on tc39-beta.netlify.com

keithamus:
Fixes #64 

> You can make them absolute which tends to make development across forks/repos easier:

> ```html
> <link href="{{site.github.url}}/css/....">
> ```

> Locally site.github.url is just ''. When served through gh-pages it'll be https://tc39.github.io/beta, and when forked to e.g. me it'll be https://keithamus.github.io/beta

> Also while on the topic of URLs - it is preferred to avoid protocol relative URLs (that is, urls without the http, e.g. src="//cdnjs.cloudflare.com...". This should be avoided because if you have an https capable asset, it should ideally always be served with https. (See https://www.paulirish.com/2010/the-protocol-relative-url/)

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -8,11 +8,11 @@
     <link rel="icon" type="image/png" href="favicon-32x32.png" sizes="32x32" />
     <link rel="icon" type="image/png" href="favicon-16x16.png" sizes="16x16" />
     <link href="https://fonts.googleapis.com/css?family=Open+Sans:300,400,600|Source+Sans+Pro:300,400,600" rel="stylesheet">
-    <link rel="stylesheet" href="/assets/css/normalize.css">
-    <link rel="stylesheet" href="{{ '/assets/css/style.css?v=' | append: site.github.build_revision | relative_url }}">
-    <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.12.0/styles/default.min.css">
-    <script defer src="{{ '/assets/js/main.js?v=' | append: site.github.build_revision | relative_url }}" type="text/javascript"></script>
-    <script src="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.12.0/highlight.min.js" type="text/javascript"></script>
+    <link rel="stylesheet" href="{{ site.github.url }}/assets/css/normalize.css">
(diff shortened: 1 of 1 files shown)


ljharb:
I'm confused about why we wouldn't use `/` (iow, root-relative URLs) which would handle inserting the right domain name in every case.

Regaddi:
The fix doesn't work for me locally 😔. It tries to access e.g. `https://github.com/pages/Regaddi/tc39-beta/assets/css/style.css?v=65d61e221df72a92b2a79d4987135c9fd8e9243f` instead of `/assets/css/style.css?v=65d61e221df72a92b2a79d4987135c9fd8e9243f`.

Do I miss some configurational part?

IgnoredAmbience:
@ljharb Root-relative URLs would fail when the repository is forked and/or renamed, for example if I wanted my fork to be named "tc39-website" instead of "beta".

ljharb:
@IgnoredAmbience interpolating the repo name makes sense to me, I’m just asking about the full url with domain - can we do just the repo name?

keithamus:
I personally don't know of a way to do the relative URL. I think the function just reads from config. Happy if someone can come up with a better solution but this is how I know how to do it.

codehag:
netlify was failing due to the repo name changed, should be fixed now

jasonwilliams:
@codehag still failing :/

jasonwilliams:
@keithamus  This isn't working locally.
links are resolving to:
`https://github.com/pages/tc39/beta/assets/js/main.js?v=7a0521117545f2971647f6d467df9062e035beef`
`https://github.com/pages/tc39/beta/assets/css/style.css?v=7a0521117545f2971647f6d467df9062e035beef`
`https://github.com/pages/tc39/beta/assets/css/style.css?v=7a0521117545f2971647f6d467df9062e035beef`

codehag:
Fixed by #72

PR #69, opened 2018-06-21
Fix #68 Move popup to outside of the window when closing it

### LINKED ISSUE #68: popup layout and animation issues
The popup obscures the content below it even though it is hidden.

The [current implementation](https://github.com/tc39/beta/blob/65d61e221df72a92b2a79d4987135c9fd8e9243f/_sass/_popup.scss#L12-L16) for hiding it hides only the `label` using the `opacity`. The links/content underneath are still obscured by the popup preventing any interactions. The parent (`div.popup`) needs to be hidden instead of the label, which might require JS.

image

And I would also like to mention that the keyframe `moving` has visually no effect because while the 90% mark to move the `label` is reached the label would have an opacity of `0.1` which makes it nearly invisible.

image

othree:
This PR make sure the popup not cover any content of the page when it is closed by move it to outside of the window.

### DIFF
--- a/_sass/_popup.scss
+++ b/_sass/_popup.scss
@@ -3,40 +3,40 @@
   left: 0;
   position: fixed;
   width: 100%;
-}
 
-input[type="checkbox"] + label {
-  position: absolute;
-  transition: 
-    opacity 0.3s ease-in-out,
-    transform 0.6s ease-in-out;
-  will-change: opacity, transform;
-}
+  input[type="checkbox"] + label {
+    position: absolute;
+    transition: 
+      opacity 0.3s ease-in-out,
+      transform 0.6s ease-in-out;
+    will-change: opacity, transform;
+  }
 
-input[type="checkbox"]:checked + label {
-  opacity: 0;
-  transform: translateY(100%);
-}
+  input[type="checkbox"]:checked + label {
+    opacity: 0;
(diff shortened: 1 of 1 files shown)


littledan:
@mram What kind of feedback were you looking for from me? I don't have an opinion on the technical aspects of this PR.

_sass/_popup.scss
```diff
@@ -6,26 +6,30 @@
 }
 
 input[type="checkbox"] + label {
```
jasonwilliams: Should this style decleration be scoped within .popup?

_sass/_popup.scss
```diff
@@ -6,26 +6,30 @@
...
-  position:relative;
+  position: absolute;
+  transition: 
+    opacity 0.3s ease-in-out,
+    transform 0.6s ease-in-out;
+  will-change: opacity, transform;
 }
 
 input[type="checkbox"]:checked + label {
```
jasonwilliams: same here

_sass/_popup.scss
r0mflip: Better to be

othree:
I added a new commit to make everything belongs to popup scoped in .popup

jasonwilliams (approved):

jasonwilliams:
Thanks!

PR #71, opened 2018-06-26
Adjust excluded files section of _config.yml

IgnoredAmbience:
A few minor tweaks to the _config.yml file to deal with which files are included/excluded in the `_site` build directory.

Exclude list:
* Add package.json and package-lock.json, NPM configuration files.
* Add screenshot.png, which is only referenced by the already-excluded
README.md.

Remove the keep_files section of the configuration. It's purpose is to
list files that should not be clobbered when the _site directory is
being rebuilt. It currently only contains README.md, and this file
is already explicitly excluded from the build by the exclude section, so
this configuration directive has no effect.

This PR also updates Gemfile.lock to the latest version of the github-pages gem (which is in-sync with the version used in production by GitHub).


codehag:
Thanks!

PR #76, opened 2018-06-29
add monthly planning call

### LINKED ISSUE #58: add information about the monthly planning call
What it says on the tin! Many people are unaware of the monthly planning call and that it is open to all. Let's add information about what kind of stuff is discussed and how to join.

codehag:
Fixes #58

I was a bit unsure if we should publish the link here, what do people think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,4 +52,4 @@ connection instructions [here](https://freenode.net/kb/answer/chat).
 ## Monthly Planning Call
 
 We have planning meetings every month, on the _last Tuesday at 10:00 AM CEST_. If you would like to
-join, come say hi at the irc channel!
+join, come say hi at the irc channel! Meeting url is https://hangouts.google.com/call/lxrlUjS_VlOcuRU6BsFBAAEI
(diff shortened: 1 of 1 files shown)


serspump4299:
Like

xtuc (approved):

littledan:
How about, say to email some address to get a link?

xtuc:
> which email would you choose? Mailing list in future?

Maybe a google group? Or I can create a mailing list on `@sauleau.com` but that looks weird.

codehag:
I don't find the email solution to be a good one. I've published my email before and received harassment and spam so I am not willing to use mine. We could set up a dedicated email account that everyone has access to, but my experience with that is that only one or two people check it. How about we publish the url, but you have to knock to get in? That way we could also address them right away

littledan:
I see, apologies for this bad suggestion. Sure, publishing the URL, knocking to get in sounds good (though I'm not a Google Meet expert and I'm not sure how to make sure that they have to do that).

Issue #70, opened 2018-06-26
Color contrast of links


marcysutton:
I noticed that links are pretty low contrast on the beta site, and they'd benefit from a more impactful color than aqua/light blue. You can check out the ratio here, it's `1.86:1` where it should be at least `4.5:1` for readability - http://contrast-ratio.com/#%2377ccbb-on-%23fefefe

Using the Chrome color picker I was able to pick a contrasting color that meets contrast guidelines: `#038067`

zoepage:
Hey @marcysutton!
Thank you for your feedback. I agree, the contrast for the text links too low.
Do you think `#009999` would also work? 
(Score is 3.46, the text is 16px + semi bold, so would fit the suggestion).

codehag:
@marcysutton would you be interested in joining our monthly call?

codehag:
issue has been addressed!

marcysutton:
@codehag that sounds interesting! Is your monthly call specifically about the tc39 website, or does it include other stuff?

codehag:
@marcysutton this call is specifically about the website, but if you are interested in other ways to get involved there are other calls as well

PR #81, opened 2018-10-12
Issue #70 Add color variable for links with good contrast

Regaddi:
I introduced an additional variable for link colors and applied the color suggested by @marcysutton in #70


codehag:
Thanks!

PR #85, opened 2018-11-06
use https protocol everywhere

chicoxyzzy:


xtuc:
Actually we could omit the protocol in every URLs in the file and add `https://` to the template, all links will be in HTTPs by default

littledan (approved):
Seems like a good incremental step, even if we could do other things in the future to use https everywhere.

PR #92, opened 2018-11-27
Add Ecma Logo

codehag:
As per our discussion in the monthly meeting, I have added the logo to the text description in the intro. If anyone has some suggestion on how to make this look good, it would be very much appreciated


zoepage:
I'll take a look

rkirsling:
My two cents:
I think the logo works fine where it is (though it could perhaps be increased in size a bit), but I think the "We are part of" text is redundant. Since the paragraph begins with "Ecma International's TC39 is...", having the Ecma International logo sit directly underneath seems like a good finishing touch on its own.

codehag:
@zoepage ping ;)

zoepage:
Looks great! :)

littledan:
Yeah, this looks great to me too. I'm glad we included the Ecma logo, but in a way that fits in well with the rest of the page (e.g., not too huge).

PR #96, opened 2019-01-17
Code samples

hemanth:
Code samples for few of the proposals.

P.S: I am having issues running the docker, pinged on IRC.

### DIFF
--- a/_data/stage3.yml
+++ b/_data/stage3.yml
@@ -67,6 +67,14 @@
     - Daniel Ehrenberg
   tests:
     - https://github.com/tc39/test262/issues/1056
+  example: >
+    const theBiggestInt = 9007199254740991n;
+
+    const alsoHuge = BigInt(9007199254740991);
+    // ↪ 9007199254740991n
+
+    const hugeButString = BigInt('9007199254740991');
+    // ↪ 9007199254740991n
   resources:
     - https://github.com/tc39/proposal-bigint
 
--- a/_data/stage3.yml
+++ b/_data/stage3.yml
@@ -238,6 +238,12 @@
     - Michael Ficarra
   tests:
     - https://github.com/tc39/test262/pull/1590
+  example: >
+    const symbol = Symbol('TC39');
+
+    console.log(symbol.description); // 'TC39'
+
+    console.log(symbol.hasOwnProperty('description')); // false
   resources:
     - https://tc39.github.io/proposal-Symbol-description/
 
--- a/_data/stage3.yml
+++ b/_data/stage3.yml
@@ -1,22 +1,3 @@
-- title: Function.prototype.toString revision
-  description: for functions defined using ECMAScript code, toString must return source text slice from beginning of first token to end of last token matched by the appropriate grammar production
-  authors:
-    - Michael Ficarra
-  champions:
-    - Michael Ficarra
-  tests:
-    - https://github.com/tc39/test262/issues/1163
-  specification: https://tc39.github.io/Function-prototype-toString-revision/
-  example: >
-    // String's parse must contains the same
-    // function body and parameter list as the original.
-
-    O.gOPD({ get a(){} }, "a").get // "function a(){}"
-
-    O.gOPD({ set a(b){} }, "a").set // "function a(b){}"
-  resources:
-    - https://github.com/tc39/Function-prototype-toString-revision
-
 - title: globalThis
   description: ECMAScript Proposal, specs, and reference implementation for globalThis
   authors:
@@ -112,7 +93,7 @@
 
       get #x() { return #xValue; }
       set #x(value) {
-        this.#xValue = value; 
+        this.#xValue = value;
         window.requestAnimationFrame(this.#render.bind(this));
       }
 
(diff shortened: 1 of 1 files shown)


_data/stage3.yml
```diff
@@ -225,6 +254,15 @@
   champions:
     - Jordan Harband
     - Kevin Gibbons
+  example: >
+    const entries = new Map([
```
ljharb: i think it might be a bit confusing to call a Map `entries`; it's certainly *iterable* for entries, but that's not exactly the same thing. maybe call it `iterableOfEntries`?

hemanth: Or, should we use `[['cat', 'dog'], ['life', 42]]`?

ljharb: That could work too - but it's also nice to show that you can do it with a Map. Maybe both examples?

littledan:
These look great! Thanks for contributing. The examples make this page much more descriptive.

Let's see if we can get reviews from the proposal champions. Do you think you could look up who each of them is and @-mention them?

_data/stage3.yml
```diff
@@ -67,6 +74,14 @@
...
   tests:
     - https://github.com/tc39/test262/issues/1056
+  example: >
+    const theBiggestInt = 9007199254740991n;
+
+    const alsoHuge = BigInt(9007199254740991);
+    // ↪ 9007199254740991n
+
+    const hugeButString = BigInt('9007199254740991');
+    // ↪ 9007199254740991n
```
littledan, line 65: For BigInt, I'd prefer we contrast the Number behavior with the BigInt behavior, and not include a use of the BigInt constructor on a Number (as that could lead people to use a buggy pattern with rounding).

littledan: cc @sarahgp

_data/stage3.yml
```diff
@@ -102,6 +143,10 @@
   champions:
     - Brian Terlson
     - Michael Ficarra
+  example: >
+    [1, [2, [3]]].flat(Infinity); // [1,2,3]
+
+    [2, 3, 4].flatMap((x) => [x, x * 2]); // [[2, 4, 3, 6, 4, 8]
```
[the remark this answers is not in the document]
littledan: This output looks good to me.

hemanth:
@littledan Thank you, I had a list of [@](https://github.com/hemanth/es-next/issues/21#issue-162612474)'s not sure if I can reuse them here.

_data/stage3.yml
hemanth: Will just rename it to `iterableOfEntries`, too many examples could be confusing?

_data/stage3.yml
```diff
@@ -155,6 +204,29 @@
...
+    var regexp = /(\w+)\W*/g;
+
+    console.log(str.matchAll(regexp));
+
+    /*
+    [
+      {
+        0: "Hello ",
+        1: "Hello"
+        index: 0,
```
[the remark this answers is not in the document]
hemanth: //cc @sebmarkbage :-)

_data/stage3.yml
sarahghp: Hm, I'm not sure what you mean @littledan. I think showing how the constructor works with a `Number` literal is pretty necessary, in general, but if you mean just for the case of this example, we could definitely use something like this instead:

```js
const x = Number.MAX_SAFE_INTEGER;
// ↪ 9007199254740991, this is 1 less than 2^53

const y = x + 1;
// ↪ 9007199254740992, ok, checks out

const z = x + 2
// ↪ 9007199254740992, wait, that’s the same as above!

const a = BigInt(Number.MAX_SAFE_INTEGER);
// ↪ 9007199254740991n, this is 1 less than 2^53, and also a BigInt

const b = a + 1n;
// ↪ 9007199254740992n, checks out, still a BigInt

const c = a + 2n
// ↪ 9007199254740993n, now that's what I want
```

littledan: In general, I am afraid that people could develop false expectations about how the BigInt constructor works, and try to use it with large number literals. I have already gotten a few "bug reports" about this. I am not sure how we could use code samples to avoid the issue; I am not an expert here.

_data/stage3.yml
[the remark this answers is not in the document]
ljharb: The output has a syntax error.

_data/stage3.yml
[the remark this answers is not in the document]
ljharb: I’m the champion of matchAll, and @mathiasbynens was the champion of named capture groups.

_data/stage3.yml
ljharb: Sounds good.

_data/stage3.yml
sarahghp: So I think this with maybe a final line showing it re-rounded? There is nothing bad about showing people how the BigInt constructor is used — not everyone is going to use literals, right?

To some extent, I think we are going to have to accept that people are going to make some mistakes as they get used to new functionality — and then it will become just a known fact about Javascript. To me that's as expected. But then, I also don't get "bug report"s.

_data/stage3.yml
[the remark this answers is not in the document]
mathiasbynens: > also I’m not sure how to illustrate better that these are not objects, but arrays with string properties

I’ve always done it [this way](https://mathiasbynens.be/notes/es-regexp-proposals#matchAll), which matches the Node.js REPL:

```
…
[
	'DEADBEEF',
	index: 19,
	input: 'Magic hex numbers: DEADBEEF CAFE 8BADF00D'
]
…
```

The only downside is that it’s not valid JavaScript syntax.

_data/stage3.yml
littledan: Yeah, I just want to emphasize the happy path. Do you see a reason why someone should use the BigInt constructor together with a Number literal? I was imagining that that was subsumed by BigInt literals, but maybe I am missing something.

_data/stage3.yml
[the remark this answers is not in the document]
hemanth: This was fixed.

hemanth:
@littledan @ljharb anything else pending on this PR?

I was planning to edit this PR to adhere to [issues#66](https://github.com/tc39/beta/issues/66#issuecomment-455041578) what say?

_data/stage3.yml
sarahghp: I could image using the `BigInt` constructor with a number literal if you were programmatically piping data into some calculations with known `BigInt`s, where the literal wouldn't really work. E.g.

```js
let nums = [4, 10, 17, 22] // pretend these come from some API you do not control
let bigFriend = 9007199254740992n

nums.map((n) => bigFriend + num) // would throw
nums.map((n) => bigFriend + BigInt(num)) // would not
```

codehag:
Hey folks, I updated this to remove stage 4 proposals

codehag:
This looks good otherwise

littledan:
LGTM

PR #101, opened 2019-02-28
new feature list

codehag:
stuff that still needs to be done:

high prio
* legend explaining what the tags are, just below the stage 3 blurb. 
* data massaging

Low prio
* potentially adding implementors, so that we have the logo of the vendor where a given proposal has been implemented (nightly, chrome, safari, etc)

### DIFF
--- a/_data/stage3.yml
+++ b/_data/stage3.yml
@@ -46,6 +46,8 @@
             });
         });
       }
+  resources:
+    - https://github.com/tc39/proposal-dynamic-import
 
 - title: Legacy RegExp features in JavaScript
   description: This is a specification draft for the legacy (deprecated) RegExp features in JavaScript, i.e., static properties of the constructor like RegExp.$1 as well as the RegExp.prototype.compile method.
--- a/_includes/intro.html
+++ b/_includes/intro.html
@@ -1,5 +1,5 @@
-<section class="intro">
-  <div class="g-row">
+<section >
+  <div class="g-row intro">
     <div class="g-col-4">
       <h2 class="section-headline">TC39</h2>
       <p>
--- a/_includes/proposals.html
+++ b/_includes/proposals.html
@@ -9,70 +9,73 @@
     </i>
     {% for proposal in site.data.stage3 %}
     <section class="featurelist__item">
-      <header class="featurelist__item__intro" tabindex="0">
-        <h4>{{ proposal.title }}</h4>
-        {% if proposal.description %}
-        <p>{{ proposal.description }}</p>
-        {% endif %}
-      </header>
-      <section class="featurelist__item__info" aria-hidden="true" tabindex="-1">
-        <div>
-          <div>
-            <h5>Status</h5>
-            <ul class="featurelist__item__status">
-              <li class="featurelist__item__owner">
-                <p>Author(s)</p>
-                {% if proposal.authors | size %}
-                <p>{{ proposal.authors | join: ", " }}</p>
-                {% else %}
-                <p>–</p>
-                {% endif %}
-              </li>
-              <li class="featurelist__item__owner">
-                <p>Champion(s)</p>
-                {% if proposal.champions | size %}
-                <p>{{ proposal.champions | join: ", " }}</p>
-                {% else %}
-                <p>–</p>
-                {% endif %}
-              </li>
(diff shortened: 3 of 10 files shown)


zoepage:
That looks brilliant already. So clean! :)

I have a few nits.
- Label colors need more contrast. Maybe we should dismiss the hover color or swap it?
- Border-bottom of navigation bar needs to go as the whole background is already grouping the navigation items.
- 3 boxes above the fold need a white background to increase contrast and readability.
- `Show Example` link in feature list should move on the right side and needs more padding-top to be more visible. It is also serving as an indicator for the examples.
- Animation for arrow of `Show Example` in feature list should change to `rotate 180°`.
- Labels (e.g. `Author`) in feature list should be bold

zoepage:
That looks brilliant already. So clean! :)

I have a few nits.
- Label colors need more contrast. Maybe we should dismiss the hover color or swap it?
- Colors of navigation highlights need to be changed to the "blue-green" we use below-
- First dropdown needs alignment with white box of navigation.
- Border-bottom of navigation bar needs to go as the whole background is already grouping the navigation items.
- 3 boxes above the fold need a white background to increase contrast and readability.
- `Show Example` link in feature list should move on the right side and needs more padding-top to be more visible. It is also serving as an indicator for the examples.
- Animation for arrow of `Show Example` in feature list should change to `rotate 180°`.
- Labels (e.g. `Author`) in feature list should be bold

codehag:
Thanks! 

With the labels -- I want to make sure they do not distract from higher importance information. Also maybe they need a hover text to explain what they are.. Maybe we can increase the opacity.

Agree with the rest

zoepage:
Another suggestion would be to remove the background color and increase the border-size ( with the full sat color) and just have the full saturation on hover.

codehag:
I addressed three of the nits
* Label colors need more contrast. Maybe we should dismiss the hover color or swap it?
* Labels (e.g. Author) in feature list should be bold
* First dropdown needs alignment with white box of navigation.

However, after I went through all of the comments and tried them, I realized that they should be moved out into separate issues, and that they need more discussion. I would like to say that we can always iterate. 

Some less contentious items:

* Show Example link in feature list should move on the right side and needs more padding-top to be more visible. It is also serving as an indicator for the examples.

Moving this to the right pulls it out of the flow of reading. I tried this while iterating on it, and I found I would frequently miss this label if I was looking for examples. I agree that it is not very visible, so I made it bold and underlined. Adding padding is a bit tricky, the goal here is to compress information. This should be iterated on in a separate issue.

This is what it looks like now:

screenshot 2019-03-04 at 13 45 42

I am not super happy about it, combined with the bold authors, because now we have a lot of weird stuff with bold going on in the text. I would prefer to use one font weight here. But I do not want to spend much more time on this and if we want to change this, precise, actionable feedback is preferred.

* Animation for arrow of Show Example in feature list should change to rotate 180°.

With regards to the animation, this is the animation we had before and I didn't change it. I would say that should we want to iterate on it, it should be moved into a separate bug. I did try it, but it did not look very good making this change and should have more time dedicated to it.

Otherwise, there are some items which, after implementing them, I realized I do not agree with or perhaps didn't understand. They should be moved out into their own issues, if we want to pursue them, however I have strong feelings against them after seeing them. 

Specifically:

* Border-bottom of navigation bar needs to go as the whole background is already grouping the navigation items.

My goal with this change was to make this stand out, removing the bottom border makes it almost blend in with the background and look unintentional:

screenshot 2019-03-04 at 12 31 36

Removing the white background likewise has a similar effect:

screenshot 2019-03-04 at 12 34 35

Since one of the major requests was that this bar stood out more, I think that it should stay as it is in this pr. At least for now, we can revisit it once we have more time. 

* 3 boxes above the fold need a white background to increase contrast and readability.

I checked the contrast and there is no readability issue. Having every text area in a white background will reduce importance of the main navigation, and makes the visual layout very busy.

Suggested change:

screenshot 2019-03-04 at 13 40 39

As an alternative we can have it all as a single white space, but I do not find that we have any real improvement in attention here. 

screenshot 2019-03-04 at 13 42 42

Let's discuss these separately

codehag:
After talking with ola offline, we agreed to merge and follow up in separate bugs

PR #102, opened 2019-02-28
Add presented

codehag:
goal of this pr is to add the last presented dates to the proposals. This can be iterated on and more proposals can be added over time.


_data/stage3.yml
```diff
@@ -219,3 +242,35 @@
...
+    #!/usr/bin/env node
+
+    // in the Script Goal
+
+    'use strict';
+
+    console.log(1);
+
+
+    #!/usr/bin/env node
```
chicoxyzzy, line 273: We should indicate that this is a first line of another script

codehag: thanks, opened a bug to have more than one script.

PR #103, opened 2019-02-28
change stage-3 blurb and title to explain stage 3

codehag:


_includes/proposals.html
```diff
@@ -2,10 +2,10 @@
   <a id="proposals" tabindex="-1"></a>
   <h2 class="section-headline text-end">State of Proposals</h2>
   <div class="featurelist">
-    <h3>Stage 3</h3>
+    <h3>Upcoming additions to the Specification</h3>
     <i class="featurelist__desc">
-      Proposals that have been reviewed, and are being considered as
-      candidates for the spec are considered to be stage 3.
+      This section features proposals that are close to completion. We call these "stage 3"
```
ljharb: “We call these” makes it sound like we don’t have a non-vague process that defines it; what’s the reason to make this more vague?

codehag: I should clarify -- the commit message is not clear.

It was mentioned by someone on the committee that "Stage 3" is meaningless to folks who are not in the committee, and just browsing the website. The goal of this blurb is to contextualize this set of proposals in some way, and inform people that there is such a thing called "stage 3" but what that means is that they are close to completion.

ljharb: Can we do that by linking to the process document, or to a different page on this website that contextualizes what the (concrete, not vague) stages mean?

I totally ack that “stage X” is meaningless to many, but that’s not the same as it being a casual nickname we use to categorize proposals.

codehag: is the issue that you are raising that stage 3 is in quotations? would the original blurb work better for you?

The stages are already contextualized above the fold, I can add a link here as well.

ljharb: It’s that the phrase “we call” connoted to me that it is an arbitrary and casual nickname, as opposed to a tightly defined stage in our process.

jmdyck: Maybe: `In [our process], these are referered to as "stage 3" proposals.`

ljharb:
```suggestion
      This section features proposals that are in Stage 3 of [our process](#link-to-process-here), which means they are close to completion and likely ready to be shipped in engines.
```
?

_includes/proposals.html
```diff
@@ -2,10 +2,10 @@
   <a id="proposals" tabindex="-1"></a>
   <h2 class="section-headline text-end">State of Proposals</h2>
   <div class="featurelist">
-    <h3>Stage 3</h3>
+    <h3>Upcoming additions to the Specification</h3>
```
rkirsling: This header text is the part that I find most concerning—it seems that we're now implicitly promising that Stage 4 will be reached. We'd at least need the word "expected", but it's probably better to maintain the "candidate" verbiage.

Here's one possibility:
```suggestion
    <h3>Current Candidates for the Specification</h3>
```

_includes/proposals.html
```diff
@@ -2,10 +2,10 @@
   <a id="proposals" tabindex="-1"></a>
   <h2 class="section-headline text-end">State of Proposals</h2>
   <div class="featurelist">
-    <h3>Stage 3</h3>
+    <h3>Current Candidates for the Specification</h3>
     <i class="featurelist__desc">
-      Proposals that have been reviewed, and are being considered as
-      candidates for the spec are considered to be stage 3.
+      This section features proposals that are in Stage 3 of [our process](#link-to-process-here), which means they are close to completion and likely ready to be shipped in engines.
```
rkirsling:
```suggestion
      This section features proposals that are in Stage 3 of <a href="https://tc39.github.io/process-document/">our process</a>, which means they are close to completion and likely ready to be shipped in engines.
```

_includes/proposals.html
```diff
@@ -2,10 +2,10 @@
...
+      proposals.
```
rkirsling: Someday we'll get multiline suggestions... 😛 
```suggestion
```

codehag:
@ljharb @rkirsling  how are we looking?

_includes/proposals.html
```diff
@@ -2,10 +2,9 @@
   <a id="proposals" tabindex="-1"></a>
   <h2 class="section-headline text-end">State of Proposals</h2>
   <div class="featurelist">
-    <h3>Stage 3</h3>
+    <h3>Current Candidates for the Specification</h3>
     <i class="featurelist__desc">
-      Proposals that have been reviewed, and are being considered as
-      candidates for the spec are considered to be stage 3.
+      This section features proposals that are in Stage 3 of <a href="https://tc39.github.io/process-document/">our process</a>, which means they are close to completion and likely ready to be shipped in engines.
```
codehag: it was mentioned that "likely ready to be shipped in engines" might be contentions. Is there another way to say this?

littledan:
I'd suggest just deleting the summary. The truth is, what Stage 3 means is a bit subtle. We are working on better documentation here as part of the how-we-work project; let's link to that when it's ready.

_includes/proposals.html
littledan: Stage 3 things are not necessarily likely to be shipped in engines. For example, SIMD.js was Stage 3 for a long time and never shipped in engines.

ljharb: Likely *ready*, which SIMD was - not likely to be shipped.

codehag: how about we remove the last part of the sentence? "This section features proposals that are in Stage 3 of our process, which means they are close to completion"

Or we can add "and likely ready for consideration as part of the specification of the language" -- its pretty vague.

littledan: SGTM (the first part)

syg:
@littledan Are you suggesting delete any mention of "Stage 3" or only the part about what stage 3 implies about the status of the proposals? I think it's fine to say these are Stage 3 without any mention of likelihood of shipping.

littledan:
@syg Ultimately, I like the suggestion of @codehag , "This section features proposals that are in Stage 3 of our process, which means they are close to completion". I would rather not characterize Stage 3 much further, as it's a complicated topic, and we can document it more in how-we-work.

syg:
That sounds great to me.

littledan (approved):

PR #112, opened 2019-03-05
fixes #109 - make links same color

### LINKED ISSUE #109: Use a single color for all links
We have two colors for links at the moment:

#7cb

and 

#017e78

lets use one -- #017e78

codehag:
There is an issue that was surfaced with this change -- now the only element that uses the color is the stage 3 header. I tried to keep it consistent but i am not sure

### DIFF
--- a/_sass/_variables.scss
+++ b/_sass/_variables.scss
@@ -1,6 +1,6 @@
 // Colors
 $color-tc39: #f90;
-$color-primary: #7cb;
+$color-primary: #017e78;
 $color-link: #017e78;
 $color-text: #000;
 $color-nav: #444;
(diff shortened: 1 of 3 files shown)


_sass/_variables.scss
```diff
@@ -1,6 +1,6 @@
 // Colors
 $color-tc39: #f90;
-$color-primary: #7cb;
+$color-primary: #017e78;
```
codehag, line 3: this should be deleted if we decide that we want to use the same color everywhere. this is open to discussion

jasonwilliams:
The new dark colour is an improvement for me 👍

jasonwilliams (approved):

PR #122, opened 2019-03-09
Improve menu

codehag:
I noticed a few issues with the menu on mobile and desktop:

Mobile
* position of hamburger menu was strange
* menu disappeared into the page 

Desktop
* jumping text when hovering (I removed the weight. An alternative is to change the background color. Trying to match the width of the text did not look much better)


codehag:
To view the preview, click "show all checks" and then "details" for deploy/netlify

Screenshot 2019-03-09 at 20 09 08

PR #125, opened 2019-03-09
fix patent agreement link

codehag:
@littledan can you take a look if i got the correct patent agreement here?


littledan:
Nope, people who join Ecma should not sign this form.

littledan:
We can just delete the patent link from here. Soon, in how-we-work, we will have a more detailed explanation of how to join

PR #137, opened 2019-03-18
Fix Scrollbar

davidbernegger:
## Problem

- When the window is too narrow, a scrollbar appears

image
- Pictured: Chrome right, Firefox left (current dev versions)

## Solution

The problem is the width of the .menu pane `100vh` and the additional padding on the right (20px) which lets the element overflow. 
 
- Either change width: 100vw to 100% (in this pr) 
- Or remove the line altogether (as right 0, left 0 negates that anyways)
- Or change to the width to calc(100vw - 20px) because of the padding.

## Solution Pictured

image
- no scrollbars anymore

### DIFF
--- a/_sass/_page-menu.scss
+++ b/_sass/_page-menu.scss
@@ -74,7 +74,7 @@ $trans-timing: 600ms;
     right: 0;
     top: 0;
     transition: max-height $trans-timing ease;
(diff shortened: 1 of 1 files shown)


littledan:
The fix makes sense. Thanks!

Issue #139, opened 2019-03-19
Orange tag has insufficient contrast


aardrian:
The orange tag for specification has insufficient contrast with its white text (see [WCAG 2.1 SC 1.4.3](https://www.w3.org/WAI/WCAG21/quickref/?showtechniques=141#contrast-minimum) and [G18: Technique for meeting 1.4.3](https://www.w3.org/TR/WCAG20-TECHS/G18.html)). The text is white (#fff) and the background is orange (#f90), producing a contrast ratio of 2.1:1, well below the 4.5:1 minimum.

![2019-03-19 (2)](https://user-images.githubusercontent.com/1376607/54600790-ee0a5780-4a13-11e9-9d0d-791692cfbd12.png)

Orange and white generally have poor contrast, so you will end up with an almost brown color if you try to darken it to meet 4.5:1.

littledan:
@aardrian Thanks for raising this issue. Would you be interested in making a PR to fix it?

aardrian:
Sure. Do you have a color palette already defined or some other parameters? Or can I just make a choice/suggestion?

littledan:
Suggestions would be welcome.

aardrian:
Fixed with https://github.com/tc39/tc39.github.io/pull/140.

PR #140, opened 2019-03-19
TC39 orange tweaked to 4.5:1 contrast ratio on white.

aardrian:
This is part of a couple changes to make the orange tags meet WCAG 2.1 contrast requirements per issue #139 .

### DIFF
--- a/_sass/_variables.scss
+++ b/_sass/_variables.scss
@@ -1,5 +1,6 @@
 // Colors
 $color-tc39: #f90;
+$color-tc39-a11y: #AB6700;
 $color-primary: #017e78;
 $color-link: #017e78;
 $color-text: #000;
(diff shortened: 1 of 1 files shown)


littledan:
This tweak makes sense to me. In particular, this orange is to be used when the text is light, whereas the other orange is to be used when the text is dark, right? I would suggest naming it tc39-orange-dark, but feel free to leave it as is if you prefer this name.

aardrian:
I changed it to `tc39-dark`. It needs to make sense to project maintainers, not me.

ExE-Boss:
New color preview: `#AB6700`

littledan:
Does the color need to be repeated rather than referencing the new variable?

aardrian:
@littledan I suspect not. I am not a SASS guy, so I defer to you on whether one variable can reference another variable. Happy to edit.

littledan:
Thanks for the fix!

PR #143, opened 2019-03-19
Duplicated hover styles to focus

aardrian:
This supports keyboard-only users by replicating the hover states as focus styles. The catch is that the hover states are on the container _and_ the interactive control, necessitating `:focus-within`. Because this selector will kill an entire style for IE users when added into another selector, it warrants its own stand-alone selector (and still leaves IE users unsupported for focus styles, so there is still a gap). Also, I do not know SASS (woo!).

### DIFF
--- a/_sass/_featurelist.scss
+++ b/_sass/_featurelist.scss
@@ -23,6 +23,9 @@
       &:hover {
         background-color: transparent;
       }
+      &:focus-within {
+        background-color: transparent;
+      }
       a {
         color: white;
         text-decoration: none;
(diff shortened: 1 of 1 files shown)


aardrian:
BTW, if you want to test in your browser dev tools, these are the styles I am trying to get into your SASS files:

```
.featurelist__item__tag:focus-within {
    background-color: transparent;
}

.featurelist__item__presented:focus-within a {
    color: #a63446;
}

.featurelist__item__tests:focus-within a {
    color: #017e78;
}

.featurelist__item__spec:focus-within a {
    color: #AB6700;
}
```

littledan (approved):

PR #145, opened 2019-03-20
Link to ECMA using https

othree:
The links to `http://www.ecma-international.org/` are still using http protocol.
The site is available through https protocol.
So update the links to https.


xtuc (approved):
Thank you, that lgtm.

PR #150, opened 2019-03-27
Move 'details' pointer cursor to only summary

larsgw:
Do not show a `pointer` cursor for the entire `details` element, only the `summary`. The current behavior is misleading, since clicking on other parts of `details` (apart from the badges) does not close the `details`.

### DIFF
--- a/_sass/_featurelist.scss
+++ b/_sass/_featurelist.scss
@@ -171,6 +171,6 @@
     text-align: center;
   }
 }
-details {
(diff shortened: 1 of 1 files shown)


rkirsling (approved):
Thanks! It took me a moment to realize that the issue is specifically when hovering over the expanded text, but this is definitely a good fix.

littledan:
Good catch.

PR #160, opened 2019-06-06
fix #156, improve keyboard navigation

### LINKED ISSUE #156: Reassess keyboard accessibility
Keyboard navigation could be improved:

- Add `Skip to content` link before navigation
- Remove tabindex from proposal sections
- As stated in #147 add tabindex to `Show Example`

Edit: Previous version of this issue relates to Firefox on Mac OS.

Regaddi:
Fixes #156.

:warning: Includes #158. Relevant changes are in 7164692 

### Notes

I removed some old JavaScript logic, which related to an older state where proposals were listed in an accordion like manner. Because the visual representation of the proposals changed, I removed this code.

I also restructured the proposals HTML: previously there were a lot of unnecessary nested `section` elements rendered. I think every proposal can be treated as an `article` instead with a `section` for the example.

I also added a `Skip to content` link, to allow to skip the navigation.

This is no necessarily a 100% solution, but should solve general issues with the page.

### DIFF
--- a/_includes/contribute.html
+++ b/_includes/contribute.html
@@ -1,7 +1,6 @@
 <section>
-  <a id="contribute" tabindex="-1"></a>
   <div class="g-row">
-    <h2 class="section-headline g-col-4">Contribute</h2>
+    <h2 id="contribute" class="section-headline g-col-4">Contribute</h2>
     <div class="g-col-8">
       <p>
         TC39 welcomes contributions from the JavaScript community, whether it is feedback on existing proposals, improved documentation,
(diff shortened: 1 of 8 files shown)


littledan:
@bkardell, could you review this?

codehag:
@Regaddi can you rebase this since 158 was merged?

bkardell:
This all looks reasonable - it's very hard to evaluate in full without actually seeing it stitched together, which I can't do atm unfortunately - but it definitely should fix a few things and doesn't seem like it makes anything worse.

littledan:
I'm just going to land this, and we can revert it if anyone reports a regression.

PR #171, opened 2019-10-16
Correct stage count

mathiasbynens:
Per the referenced process document, there are five stages (not four):

1. Stage 0
2. Stage 1
3. Stage 2
4. Stage 3
5. Stage 4

### DIFF
--- a/_includes/intro.html
+++ b/_includes/intro.html
@@ -34,7 +34,7 @@
         We develop the JavaScript (formally, ECMAScript) specification
         <a href="https://github.com/tc39/ecma262/">on GitHub</a> and meet every two
         months to discuss proposals. To learn more about the process, please take a look at the
-        <a href="https://tc39.es/process-document/">four stages</a> for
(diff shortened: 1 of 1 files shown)


littledan:
I don't think Stage 0 is really much of a stage, even if we have documented the term in the process doc. I prefer to continue to refer to it as a four-stage process.

jasonwilliams:
> I don't think Stage 0 is really much of a stage, even if we have documented the term in the process doc. I prefer to continue to refer to it as a four-stage process.

Does that not make this awkward?
https://github.com/tc39/proposals/blob/master/stage-0-proposals.md

ljharb:
@rkirsling hence the two bullet points on the linked document, filtering the list from “uselessly broad” to “useful”

bramus:
The [Process Document](https://tc39.es/process-document/) also states that there are five stages:

> There are five stages: a strawperson stage, and 4 “maturity” stages.

mathiasbynens:
I gotta say, I wasn't expecting this to be controversial...

PR #181, opened 2019-12-16
Correct the amount of stages

bramus:
The site mentions "four stages", yet it's a five stage process _(as detailed in [the Process Document](https://tc39.es/process-document/))_.

This PR fixes that error.


bramus:
Oops, duplicate of #171 

¯\_(ツ)_/¯

PR #185, opened 2020-01-23
Fixes incorrect description for Promise.any proposal

lourd:
Followed the style from the Promise.allSettled description

### DIFF
--- a/_data/stage3.yml
+++ b/_data/stage3.yml
@@ -474,7 +474,7 @@
 - title: String.prototype.replaceAll
   id: proposal-string-replaceall
   has_specification: true
-  description: introduce String.prototype.replaceAll
+  description: Introduce String.prototype.replaceAll as a way to replace all in a string
   authors:
     - Peter Marshall
     - Jakob Gruber
@@ -492,7 +492,7 @@
 - title: Promise.any
   id: proposal-promise-any
   has_specification: true
-  description: Introduce String.prototype.replaceAll as a way to replace all in a string
+  description: This proposal is to match popular usage in libraries, adds a method which returns a promise that is fulfilled by the first given promise to be fulfilled, or rejected if all of the given promises are rejected.
   authors:
     - Mathias Bynens
     - Kevin Gibbons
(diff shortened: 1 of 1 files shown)


_data/stage3.yml
```diff
@@ -492,7 +492,7 @@
 - title: Promise.any
   id: proposal-promise-any
   has_specification: true
-  description: Introduce String.prototype.replaceAll as a way to replace all in a string
+  description: This proposal is to match popular usage in libraries, adds a method which returns a promise that is fulfilled by the first given promise to be fulfilled, or rejected if all of the given promises are rejected.
```
ljharb: the first clause (which needs a conjunction if kept) doesn't seem to add anything here to me.
```suggestion
  description: Adds a method which returns a promise that is fulfilled by the first given promise to be fulfilled, or rejected if all of the given promises are rejected.
```

ljharb (approved):

_data/stage3.yml
lourd: Agreed!

_data/stage3.yml
```diff
@@ -492,7 +492,7 @@
 - title: Promise.any
   id: proposal-promise-any
   has_specification: true
-  description: Introduce String.prototype.replaceAll as a way to replace all in a string
+  description: Adds a method which returns a promise that is fulfilled by the first given promise to be fulfilled, or rejected if all of the given promises are rejected.
```
SamB: Well, I was mentally nitpicking whether this was a method or a function, but then I realized it didn't really matter. 

A bit later, I realized that if it doesn't matter, it might as well get trimmed from the page, too:

```suggestion
  description: Returns a promise that is fulfilled by the first given promise to be fulfilled, or rejected if all of the given promises are rejected.
```

I won't lose any sleep if you don't like this change, though.

SamB (approved):
Looks pretty good to me.

_data/stage3.yml
lourd: Haha, I appreciate your conviction for brevity. I think technically it is a static method, though, so I'm going to leave the extra words.

codehag:
Thanks for this

PR #189, opened 2020-02-10
Fix #149, Tag title and link should work in entire element

### LINKED ISSUE #149: Move tag hover description from text to whole tag
Currently the hover description only shows when hovering over the text in the tag, not the entire tag.

See #107

othree:
Fix #149. Make the `<a>` display block and move the tag style to the `<a>` element. Which will make the anchor element fill the entire tag. This change solves:

1. The hover tip(title) works in entire tag element, this is issue #149 
2. Click the link works in entire tag element(used to only work when click on the link text)

Not a perfect fix because I have to disable some stylelint rule in certain line. But this change is small. A better solution will have to modify the HTML markup structure. For example, remove the `<li>` element of tag list.

### DIFF
--- a/_sass/_featurelist.scss
+++ b/_sass/_featurelist.scss
@@ -15,21 +15,23 @@
     }
     &__tag {
       display: inline-block;
-      font-size: 0.8rem;
-      margin-top: 0.5rem;
-      opacity: 1;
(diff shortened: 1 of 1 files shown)


codehag:
Looks good, thanks

PR #201, opened 2020-07-15
fix #192, add missing footer

### LINKED ISSUE #5: Footer with links

### LINKED ISSUE #192: Copyright Notice Missing From Footer

DerekNonGeneric:
Notes:
&emsp;https://user-images.githubusercontent.com/17770407/87853458-340d4f00-c8d8-11ea-889b-b8c476f7c3cb.png
&emsp;https://user-images.githubusercontent.com/17770407/87853471-4ab3a600-c8d8-11ea-8484-71242942ec23.png
&emsp;https://www.color-hex.com/color/696969
&emsp;https://webaim.org/resources/contrastchecker/?fcolor=696969&bcolor=fafafa&api

Fixes: #192

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -16,15 +16,15 @@
     <script>hljs.initHighlightingOnLoad();</script>
   </head>
   <body class="no-js">
-    <main class="main">
     {% include_relative _layouts/page_header.html %}
+    <main class="container">
     <h1 id="content" class="page-title">Specifying JavaScript.</h1>
     <section class="main-content">
       {% include intro.html %}
       {% include proposals.html %}
     </section>
-    {% include_relative _layouts/page_footer.html %}
     </main>
+    {% include_relative _layouts/page_footer.html %}
     <img class="hidden" src="https://d3nrsflo8ae6b5.cloudfront.net/1x1.png" alt="pixel" />
   </body>
 </html>
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -23,18 +23,8 @@
       {% include intro.html %}
       {% include proposals.html %}
     </section>
+    {% include_relative _layouts/page_footer.html %}
     </main>
-
-    <div class="popup">
-      <input type="checkbox" id="one" class="hidden" name="ossm">
-      <label for="one" class="alert-message">
-        <strong>We are
-        looking for feedback.</strong> Please add issues to our <a
-                    href="https://github.com/tc39/tc39-web-draft/issues" target="_blank">GitHub
-                    page</a>.
-        <span class="close">×</span>
-      </label>
-    </div>
-    <img src="https://d3nrsflo8ae6b5.cloudfront.net/1x1.png" alt="pixel" />
+    <img class="hidden" src="https://d3nrsflo8ae6b5.cloudfront.net/1x1.png" alt="pixel" />
   </body>
 </html>
--- a/_layouts/page_footer.html
+++ b/_layouts/page_footer.html
@@ -1,0 +1,15 @@
+<footer class="page-footer">
+  <div class="grid-container">
+    <div class="grid-start">
+      <span>We are looking for feedback.<br />Please <a href="https://github.com/tc39/tc39.github.io/issues/new" target="_blank">open an issue</a> using the issue tracker for <a href="https://github.com/tc39/tc39.github.io" target="_blank">our site repository</a>.</span>
+    </div>
+    <div class="grid-end">
+      <ul class="pipe-nav">
+        <li><a href="https://tc39.es/archives/">Archives</a></li>
+        <li><a href="https://tc39.es/process-document/">Process</a></li>
+        <li><a href="https://tc39.es/code-of-conduct/">Code of Conduct</a></li>
+      </ul>
+      <span>&copy; Ecma International</span>
+    </div>
+  </div>
+</footer>
--- a/_layouts/page_footer.html
+++ b/_layouts/page_footer.html
@@ -1,4 +1,4 @@
-<footer class="page-footer">
+<footer class="page-footer container">
   <div class="grid-container">
     <div class="grid-start">
       <span>We are looking for feedback.<br />Please <a href="https://github.com/tc39/tc39.github.io/issues/new" target="_blank">open an issue</a> using the issue tracker for <a href="https://github.com/tc39/tc39.github.io" target="_blank">our site repository</a>.</span>
--- a/_layouts/page_footer.html
+++ b/_layouts/page_footer.html
@@ -3,7 +3,7 @@
     <div class="grid-start">
       <span>We are looking for feedback.<br />Please <a href="https://github.com/tc39/tc39.github.io/issues/new" target="_blank">open an issue</a> using the issue tracker for <a href="https://github.com/tc39/tc39.github.io" target="_blank">our site repository</a>.</span>
     </div>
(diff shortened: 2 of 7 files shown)


DerekNonGeneric:
I decided not to tackle #5 in this PR, but got pretty far along before realizing that it would probably be best to split them up. If these changes look good to go and are merged (you may need to squash it first), I'll open up my PR that fixes #5. Those changes are a good segue into talking about the information architecture, which may need to be re-evaluated to achieve the desired automation.

/cc @codehag @ljharb

_layouts/page_footer.html
```diff
@@ -0,0 +1,15 @@
+<footer class="page-footer">
+  <div class="grid-container">
+    <div class="grid-start">
+      <span>We are looking for feedback.<br />Please <a href="https://github.com/tc39/tc39.github.io/issues/new" target="_blank">open an issue</a> using the issue tracker for <a href="https://github.com/tc39/tc39.github.io" target="_blank">our site repository</a>.</span>
```
DerekNonGeneric, line 4: I changed this from &hellip;

> Please add issues to our GitHub page.

&hellip; to &hellip;

> Please open an issue using the issue tracker for our site repository.

_layouts/default.html
```diff
@@ -23,18 +23,8 @@ <h1 id="content" class="page-title">Specifying JavaScript.</h1>
       {% include intro.html %}
       {% include proposals.html %}
     </section>
+    {% include_relative _layouts/page_footer.html %}
```
codehag: should this be in the main section?

_sass/_page-footer.scss
```diff
@@ -0,0 +1,66 @@
...
+}
+
+.page-footer a {
+  color: $color-eclipse;
+  font-weight: 400;
+  text-decoration: none;
+}
+
+.page-footer a:hover {
+  color: $color-eclipse;
```
codehag: I noticed that when looking at it, it wasn't super obvious to me that these are links. what do you think about making this the link color? https://github.com/tc39/tc39.github.io/blob/master/_sass/_variables.scss#L5

codehag (commented):
Looks good, a few nits

_layouts/default.html
DerekNonGeneric: It's a good question. I decided to mirror the placement of the pre-existing `page_header.html` above it since it's only a single page for now. https://github.com/tc39/tc39.github.io/blob/master/_layouts/default.html#L20

For a multi-page site, I would go w/ more of a so-called &#8220;global footer&#8221; that would exist outside of the [`<main>` element](https://github.com/tc39/tc39.github.io/blob/master/_layouts/default.html#L19), which would have started at [`default.html#L27`](https://github.com/tc39/tc39.github.io/blob/master/_layouts/default.html#L27) instead.

_sass/_page-footer.scss
DerekNonGeneric: I'm really glad you brought up color.

I was initially going for a so-called &#8220;muted&#8221; aesthetic for the footer, but I like the idea of making the links that color. As you may already know, to be technically correct here (in the context of web a11y), the link color needs to have a contrast ratio > 4.5 to pass AA. I checked, and if we were to use that color with the existing background color, we wouldn't be passing [axe](https://chrome.google.com/webstore/detail/axe-web-accessibility-tes/lhdoppojpmngadmnindnejefpokejbdd) because (for some odd reason) it rounds the `#fafafa` to `#f0f0f0`.

Proof: https://webaim.org/resources/contrastchecker/?fcolor=017e78&bcolor=f0f0f0&api

_sass/_variables.scss
```diff
@@ -8,6 +8,9 @@ $color-nav: #444;
 $color-spacer: #eee;
 $color-submenu-bg: white;
 $color-bg: #fafafa;
```
DerekNonGeneric, line 10: Continuation of https://github.com/tc39/tc39.github.io/pull/201#discussion_r456735261&hellip;

To be technically correct here, while passing AA, we are going to need to change the background color from `#fafafa` to `#f5f5f5`, which would allow us to use that link foreground while passing.

Proof: https://webaim.org/resources/contrastchecker/?fcolor=017e78&bcolor=f5f5f5&api

```suggestion
$color-bg: #f5f5f5;
```

_sass/_variables.scss
```diff
@@ -8,6 +8,9 @@ $color-nav: #444;
...
+$color-eclipse: #3d3d3d;
+$color-gray59: #969696;
```
DerekNonGeneric: As for the color of the rest of the text, the most-muted (least-contrasting) color (with repeating two-digit hex pair) we can use is `#737373`.

https://www.color-hex.com/color/737373
https://webaim.org/resources/contrastchecker/?fcolor=737373&bcolor=fafafa&api

```suggestion
$color-gray45: #737373;
```

DerekNonGeneric: However, the nearby-named Dim Gray may be preferable.

https://www.color-hex.com/color/696969
https://webaim.org/resources/contrastchecker/?fcolor=696969&bcolor=fafafa&api

```suggestion
$color-dim-gray: #696969;
```

image

---

Update in https://github.com/tc39/tc39.github.io/pull/201/commits/efe562c5c4efbaafc756904b3718430835b67838:

image

---

Update in https://github.com/tc39/tc39.github.io/pull/201/commits/0c4077c340a05308b78364490dc554d74ac91353:

image

DerekNonGeneric: image

_layouts/default.html
codehag: I don't think that, because it is a single page, that the footer should be inside the main content. The purpose of the structure of `<nav></nav><main></main><footer></footer>` isn't for single pages vs multi-page sites -- its for semantic html. This is used by accessibility technologies and also helps keep the site well organized -- regardless of if it is a multi page site or a single page site 

You can read more about that here https://developer.mozilla.org/en-US/docs/Learn/Accessibility/HTML and here https://webflow.com/blog/html5-semantic-elements-and-webflow-the-essential-guide

Let's move the footer out of the main body and, if you like -- in a separate patch deal with the nav. Looking at the code  -- the nav is unnecessarily nested. There is likely cleanup there. There will probably be issues in the styling due to this nesting. That might be a nice next bug.

DerekNonGeneric: It's hard to read because it isn't indented, but the page header is currently nested within the `<main>` element.

codehag: I'm aware that the header, which contains the nav, is currently in the main section. 

If you read my comment again, I proposed moving the nav in the next patch. It is nested inside of the header in a single element, which in turn is in the main section.

>  if you like -- in a separate patch deal with the nav. Looking at the code -- the nav is unnecessarily nested

Please move the footer out of the main section. There is no reason for it to be there.

DerekNonGeneric: > [&hellip;] in a separate patch deal with the nav.

Do you think it would alright if this was taken care of in https://github.com/tc39/tc39.github.io/pull/201/commits/050462461b7aca40ef96d956eec9ed20c3ebc674?

DerekNonGeneric:
Thanks for the reviews so far, I think all feedback has been incorporated. 😄

_layouts/default.html
codehag: sgtm

PR #206, opened 2020-07-29
fix #205, eliminate mobile navbar space

### LINKED ISSUE #205: Navigation space on mobile
When viewing this site on mobile devices, it is possible to scroll horizontally, which doesn't seem to be intentional. If one does so, additional space to the right of the navigation bar is visible. If one does not do so, the navigation bar's hamburger menu button appears to be lacking necessary padding. This cosmetic defect can be observed in the screenshots below, which is reproducible when using Firefox for Android. The horizontal scrolling is not visible when one zooms out, however, I do not believe users should have the ability to zoom _out_. ~~If they are, the initial viewport zoom behavior may need to be adjusted.~~

Notes:
&emsp;https://user-images.githubusercontent.com/17770407/88852310-184d5700-d1bc-11ea-999a-a86626d8fa1b.png
&emsp;https://user-images.githubusercontent.com/17770407/88852313-18e5ed80-d1bc-11ea-99ae-edf38948e158.png

P.S. https://github.com/tc39/tc39.github.io/pull/201 did not introduce this bug. Apparently it has been present for [a while](https://web.archive.org/web/20200506110801/https://tc39.es/).

DerekNonGeneric:
Notes:
&emsp;https://user-images.githubusercontent.com/17770407/88860772-7e40db00-d1ca-11ea-91c0-173b2bf72d6d.png
&emsp;https://user-images.githubusercontent.com/17770407/88860773-7ed97180-d1ca-11ea-9c89-7260cfaa8c5f.png

Fixes: https://github.com/tc39/tc39.github.io/issues/205

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -2,7 +2,7 @@
 <html lang="{{ site.lang | default: "en-US" }}">
   <head>
     <meta charset="UTF-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
     <meta name="theme-color" content="#157878">
     <title>TC39 – Specifying JavaScript.</title>
     <link rel="icon" type="image/png" href="favicon-32x32.png" sizes="32x32" />
--- a/_sass/_page-header.scss
+++ b/_sass/_page-header.scss
@@ -3,7 +3,6 @@
   background: inherit;
   display: flex;
   flex-direction: row;
-  padding: 0;
   @media (min-width: $bp-tablet) {
     $spacing: 30px;
     .page-menu {


_layouts/default.html
```diff
@@ -2,7 +2,7 @@
 <html lang="{{ site.lang | default: "en-US" }}">
   <head>
     <meta charset="UTF-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
```
DerekNonGeneric, line 5: This PR is also removing users' ability to zoom _out_ (beyond the initial scale of 1) as this should never be necessary, however, they will still be able to zoom _in_ if they choose to. More information about the topic can be found elsewhere.

> minimum-scale | The minimum amount the visitor can zoom on the page. 1.0 does not zoom.
> -- | --
>
> &mdash; https://css-tricks.com/snippets/html/responsive-meta-tag/

As further clarification that this is not a mistake, see the following quote from MDN.

> Usage of `user-scalable=no` can cause accessibility issues to users with visual impairments such as low vision.
>
> &mdash; https://developer.mozilla.org/en-US/docs/Mozilla/Mobile/Viewport_meta_tag#Viewport_basics

This means that we're still good as far as a11y is concerned.

codehag (approved):
LGTM

PR #213, opened 2020-10-12
🏗️✨ introduce ESLint

DerekNonGeneric:
To lint a lot of formats, this PR introduces ESLint. Opening this PR early (and as a draft) in hopes to solicit feedback.

No plans on making any of the corrections until it seems like these rules sound good to adopt for now.

Impossible to be unopinionated, but didn't adopt some of the larger well-known presets, so things are likely missing from this.

/cc @ljharb @codehag


.config/eslint.mjs
```diff
@@ -0,0 +1,573 @@
+// -----------------------------------------------------------------------------
```
ljharb: this is quite a large file - it seems like it'd be much easier to maintain both with the prettier config inlined, rather than in a separate prettier config file, but also by using one of the standard presets available rather than defining every single rule.

(my personal preference is also that config is static, ie, not in a .js file, but i suspect not everyone shares this view)

DerekNonGeneric: > this is quite a large file - it seems like it'd be much easier to maintain both with the prettier config inlined, rather than in a separate prettier config file

I don't want to mix them in case there are some editor plugins that look for:
- `.prettierrc`
- `.prettierrc.js`
- `.prettierrc.json`.

> but also by using one of the standard presets available rather than defining every single rule.

Are there any presets I should look into? I haven't really been shopping around.

> (my personal preference is also that config is static, ie, not in a .js file, but i suspect not everyone shares this view)

Line #388 of this file basically made the decision for me, but I also make good use of it here. I really like it myself.

ljharb: The intention is to actively prevent editors that intend to invoke prettier directly; prettier should only be ran on JS via eslint.

.config/eslint.mjs
```diff
@@ -0,0 +1,573 @@
...
+
+    // ------------------
+    // Node Plugin Rules
+    // ------------------
+
+    'node/file-extension-in-import': [OFF, 'always', { '.js': 'never' }], // TODO
+    'node/no-unsupported-features/es-syntax': [
+      ON,
+      {
+        version: pjsonObj.engines.node,
```
ljharb: this isn't necessary to specify, by default it reads from "engines" in package.json per [the docs](https://github.com/mysticatea/eslint-plugin-node/blob/HEAD/docs/rules/no-unsupported-features/es-syntax.md#configured-nodejs-version-range)

DerekNonGeneric: TIL

.config/eslint.mjs
DerekNonGeneric: > prettier should only be ran on JS via eslint

Yup, I was thinking more about for editor plugins that use it for formatting Markdown (and other) files.

DerekNonGeneric: Alright, I'm willing to combine the two files since I don't really like using Prettier for formatting Markdown.

Yulia mentioned that I may want to consider taking a look at what @mozilla uses.

https://dxr.mozilla.org/mozilla-central/source/tools/lint/eslint/.eslintrc.js

They've also got a couple of plugins that may prove useful (completely slipped my mind).

- https://dxr.mozilla.org/mozilla-central/source/tools/lint/eslint/eslint-plugin-mozilla
- https://dxr.mozilla.org/mozilla-central/source/tools/lint/eslint/eslint-plugin-spidermonkey-js

.config/eslint.mjs
DerekNonGeneric: I'm thinking about leaving it like this in case we decide to use `devEngines` in the future.

ljharb: I'm not aware that that's a thing, but also for something not published, what purpose would the distinction serve?

DerekNonGeneric: For something not published, all `devDependencies` would use `devEngines`, but you're probably right that the distinction is pointless in this situation, so nevermind. I'm going to get rid of this whole variable importing too.

.config/prettier.mjs
```diff
@@ -0,0 +1,37 @@
...
+  bracketSpacing: true,
+  endOfLine: 'lf',
+  printWidth: 80,
+  tabWidth: 2,
+  useTabs: false,
+  singleQuote: true,
+  overrides: [
+    {
+      files: [
+        '*.cjs',
```
codehag: Why do we have cjs files here?

_tools/eslint-rules/no-assign-mutated-array.mjs
```diff
@@ -0,0 +1,233 @@
+/**
+ **Derivative work of `no-assign-mutated-array.ts` by Nicholas Jamieson.
+ * @file Ban direct assignment of arrays mutated by impure array mehods.
```
codehag: I don't think this is necessary. This isn't a convention in this codebase.

tsconfig.base.json
```diff
@@ -0,0 +1,33 @@
+{
+  "compilerOptions": {
```
codehag: Do we need a typescript config? We aren't using a Typescript compiler, this will never be run.

tsconfig.eslint.json
```diff
@@ -0,0 +1,12 @@
+{
```
codehag: This file can also go...

codehag:
I should have stepped in earlier. I am a bit concerned that this is bringing so many conventions which were not a part of this codebase before. We mostly do not need this work. If you want to add eslint, can you add it as the default, and we iterate from there?

Additionally, the real benefit here will be a ~github action~ travis config.

_tools/eslint-rules/no-assign-mutated-array.mjs
DerekNonGeneric: The point of having this lint rule here is to serve as an example of what having a codebase is like that has custom rules in 2020. The rule itself is still a draft, please be patient, but it lints for [a common pitfall in JavaScript](https://twitter.com/_jayphelps/status/1196249744902967296) and if I could provide any codebase with one custom rule it would probably be this one. The unique thing about this rule is that it's an ES Module and leverages the TypeScript API (extremely rare). This pattern is very new, which is why it's still unpolished, but I ask that you please have a little bit of patience. It's almost done!

.config/prettier.mjs
DerekNonGeneric: This is simply here to cover all of our bases when trying to determine which files should be treated as JS files.

tsconfig.eslint.json
DerekNonGeneric: This is the tsconfig file for [TypeScript ESLint](https://github.com/typescript-eslint), the reason why there are two of these files is because we need a way for TypeScript ESLint to know which files it is supposed to lint while not imposing on the TypeScript compiler's config file &mdash; it extends the base for linting purposes. Believe it or not, this is actually the way that TypeScript ESLint recommends doing it and they even do it themselves in [their own repository](https://github.com/typescript-eslint/typescript-eslint).

tsconfig.base.json
DerekNonGeneric: We need it because ESLint is using the TypeScript parser. You may be wondering why one would use the TypeScript parser. The reason is because we are leveraging several TypeScript lint rules to lint our JS. This does not imply that we intend to have any TypeScript sourcecode in this codebase, but rather that we are leveraging all of the code quality tooling of that ecosystem without necessarily having to author our modules in that language (but the option is still on the table).

codehag: I took a look at the eslint repository, and this isn't the case. It is a special config for typescript repositories. Typescript is a potential extension. See https://github.com/eslint/eslint#what-about-experimental-features. I recommend taking a look at the repository, its pure js. 

Additionally, the config readme does not recommend this: https://eslint.org/docs/user-guide/configuring -- nowhere is the tsconfig mentioned.

I am really confused why we are over specifying things so much?

tsconfig.eslint.json
codehag: That is for the Typescript eslint extension.

DerekNonGeneric:
> I am a bit concerned that this is bringing so many conventions which were not a part of this codebase before.

It may seem that way, but in reality, I don't think we have a diversity problem yet. I can explain.

> Additionally, the real benefit here will be a github action.

Yes, that is in the works. Please let me know when you can get that sample you wanted to provide me with.

_tools/eslint-rules/no-assign-mutated-array.mjs
codehag: given that we have [no uses of sort anywhere in this repository](https://github.com/tc39/tc39.github.io/search?q=sort), this rule is not warranted. There is no reason to add rules like this, as they are just noise. Please only add things as necessary to the codebase, otherwise it will be confusing for new maintainers. 

Additionally, in the case we need a new plugin: if an existing, maintained plugin exists, please use that instead of writing one that we would need to maintain. We don't need anything new or experimental for this. It is just linting.

This change makes no sense here. I am getting frustrated that my comments are being ignore and argued with.

DerekNonGeneric: Sorry, I didn't mean to come off as sounding argumentative. You are right though, if you want to keep things in this repo as simple as possible, then custom rules will have to come from elsewhere. Either way, your feedback has been incorporated.

tsconfig.eslint.json
DerekNonGeneric: Yes, the TypeScript ESLint extension can lint JS code. The reason why this works out is that TS is _supposed_ to be a gradually typable superset of JS, so many of the best practices that it enforces are applicable to JS. I wish we could've gone over this PR yesterday, it's quite a fascinating topic.

tsconfig.base.json
DerekNonGeneric: Yes, I've seen the ESLint repo's codebase, it's very nice and worth revisiting. Thanks for reminding me of that.

The reason why things are overspecified and we can't just choose one of those very popular presets is because it would be extremely weird for TC39, the organization that specifies JS, to simply adopt Airbnb's or Google's code conventions since it would look like an endorsement for that company. At least that's how I see it. Maybe Jordan can clarify a bit better.

DerekNonGeneric:
> I am a bit concerned that this is bringing so many conventions which were not a part of this codebase before.

Sorry, I must've misinterpreted what you said. If you were interested in expanding the site w/ localization and things, I was thinking big. That requires quite a bit of engineering work to get done and I was trying to save you the effort of all this config.

tsconfig.eslint.json
ljharb: It's not actually one tho; it seems very strange to be using the TS parser on "not TS".

tsconfig.base.json
DerekNonGeneric: ESLint init gives three choices:

image

This config in this PR is actually using standard (actually semi-standard).

I'm going to have to see if I can trim this config file down to not override any of that style guide's rules.

ljharb: Out of those three choices, you can look at github stars and npm downloads to see the relative popularity, which I'm pretty sure matches the order presented. I'm biased, but I'd suggest the first choice.

tsconfig.eslint.json
DerekNonGeneric: What were you thinking for the parser then? I used to use `babel-eslint`, but then you can't make use of TypeScript ESLint rules (or can you?) and you need to configure Babel too, which is no fun.

DerekNonGeneric: > TS is supposed to be a gradually typable superset of JS

Sorry, to clarify what I meant with this is that the places where it diverges (array index signatures and symbols behavior) need to be avoided to make proper use of this parser in pure JS.

ljharb: It’s regular JavaScript - why not the core parser? If we need stage 3 features, babel-eslint (which can be configured with preset-env and nothing else) should be fine. Configuring babel is easier than configuring TS :-)

.config/eslint.mjs
```diff
@@ -0,0 +1,581 @@
...
+    'template-curly-spacing': ON,
+    'unicode-bom': ON,
+    'use-isnan': ON,
+    'valid-typeof': ON,
+
+    // -------------------------------
+    // TypeScript ESLint Plugin Rules
+    // -------------------------------
+
+    '@typescript-eslint/array-type': [ON, {
```
DerekNonGeneric: @ljharb, these would basically be the only rules that I use that need the TypeScript ESLint parser.

I'm not trying to be difficult. It's possible that there is another way to do this that doesn't need to use this parser.

I'm hoping that these can all be replaced with some other plugin so we can use the parser you suggest.

---

There is so little JavaScript in this repo right now, that basically the majority of it is going to be for linting the crawler, which I am in the process of developing. We don't really have to agree, but for my own purposes, these are my preferences. I will try to trim it down to bare necessities within the coming days. We may end up not needing the majority of these anyway. 😂

DerekNonGeneric: There is so little JavaScript in this repo right now, that basically the majority of it is going to be for linting the crawler, which I am in the process of developing. We don't really have to agree, but for my own purposes, these are my preferences. I will try to trim it down to bare necessities within the coming days. We may end not needing the majority of these anyway. 😂

tsconfig.eslint.json
DerekNonGeneric: I hope to have clarified the above in https://github.com/tc39/tc39.github.io/pull/213#discussion_r505023589.

.config/eslint.mjs
ljharb: We definitely don't have to agree :-) i don't make the decisions on this repo either, to be clear.

If @codehag and others think it's fine to use jsdoc, and also think the tooling complexity to verify it is worth it, then your only real options are to do this, or to use `tsc` instead of eslint to typecheck the jsdoc separately (eslint-plugin-react [does this](https://github.com/yannickcr/eslint-plugin-react/blob/master/tsconfig.json)).

tsconfig.base.json
codehag: "Standard" isn't what you want to use... The group that made it just named it "standard", it doesn't have anything to do with standards or industry agreement. Please see their readme: https://github.com/standard/standard#but-this-isnt-a-real-web-standard -- additionally, the practices that they encourage, such as leaving out semicolons, means that parsers do double work when parsing ambiguous statements.

The eslint config here is not quite right. Its a mix between the website and your crawler. We don't use esm on the website, or typescript. For the crawler it also doesn't fit either, as your crawler doesn't run in the browser. Given the simplicity of the website code, we don't need a styleguide. 

you can have multiple eslint configs within the same project. A base config for the website, which should have no options beyond the default. Then for your crawler you can make a separate config, which will be defaulted to by the directory structure. These are the options you want to use:

```
Website config:
✔ How would you like to use ESLint? · problems
✔ What type of modules does your project use? · none
✔ Which framework does your project use? · none
✔ Does your project use TypeScript? · **No** / Yes
✔ Where does your code run? · browser
✔ What format do you want your config file to be in? · JSON
Local ESLint installation not found.
The config that you've selected requires the following dependencies:
```

The resulting config file, which is the only file we need in this pr, is this:

.eslintrc.json 
```
{
    "env": {
        "browser": true,
        "es2021": true
    },
    "extends": "eslint:recommended",
    "parserOptions": {
        "ecmaVersion": 12
    },
    "rules": {
    }
}
```

That's it. It should be in the top level directory of this project. The only other changes should be npm modules that are added, total 3 files changed. Nothing else is necessary. This is all I expect from this pull request. 

Ok, for your crawler, you can create an eslint config file specific to it, and put it in the folder with your crawler. ESLint is cascading, and this config will override the base config. You can read more here: https://eslint.org/docs/user-guide/configuring#configuration-cascading-and-hierarchy. Then, when you are doing work on your crawler code, eslint will fall back on the nearest eslint config, which will be the one you specify. For this, please do it separate from this pull request -- and make it part of the pull request for your crawler.

.config/eslint.mjs
codehag: As I've mentioned elsewhere, I don't think that the tooling complexity that this brings is worth it. If you want to use these tools during your work on the crawler, its fine -> you can create a dedicated eslint config for that folder that can be reviewed with those changes. For the website however, this is too much and we should use defaults. The website should have minimal JS code as it is a static content website.

tsconfig.eslint.json
codehag: The recommended plugin for verifying JSDoc is https://github.com/gajus/eslint-plugin-jsdoc

Please see the end of life documentation for eslint's builtin JSDoc support: https://eslint.org/blog/2018/11/jsdoc-end-of-life#suggested-replacement

The typescript eslint config makes sense when you are using typescript with eslint, as the two tools use different ASTs, and purpose of the plugin is to bridge the two. See here: https://github.com/typescript-eslint/typescript-eslint#why-does-this-project-exist

We don't make use of the TypeScript AST in any capacity, so it doesn't make sense for us.

.config/eslint.mjs
```diff
@@ -0,0 +1,581 @@
...
+
+export default {
+  env: {
+    es6: true,
+    node: true,
+  },
+  extends: [
+    // Order matters; don't alphasort. Rules of later presets override previous.
+    'eslint:recommended',
+    'semistandard',
```
DerekNonGeneric, line 67: > [&hellip;] the practices that they encourage, such as leaving out semicolons, means that parsers do double work when parsing ambiguous statements.
> &mdash; https://github.com/tc39/tc39.github.io/pull/213/files#r505296951

@codehag, I did not know this, but I learned Java first and prefer the semicolons for clarity.

Omitting semicolons is also a bad idea for many other reasons too.

In the comment above I stated the following&hellip;

> This config in this PR is actually using standard (actually semi-standard).

When I said "semi-standard", I messed up and should've said "semistandard".

Sorry, still not too familiar w/ it, but here's what that is:

> All the goodness of standard/standard with semicolons sprinkled on top.
> &mdash; https://github.com/standard/semistandard

DerekNonGeneric: > [&hellip;] leaving out semicolons, means that parsers do double work when parsing ambiguous statements.
> &mdash; https://github.com/tc39/tc39.github.io/pull/213/files#r505296951

@codehag, I did not know this, but I learned Java first and prefer the semicolons for clarity.

In my comment, I stated the following&hellip;

> This config in this PR is actually using standard (actually semi-standard).
> &mdash; https://github.com/tc39/tc39.github.io/pull/213/files#r505296951

When I said "semi-standard", I messed up and should've spelled it as "semistandard".

This was a mistake (still not too familiar w/ it), but here's what that is:

> All the goodness of standard/standard with semicolons sprinkled on top.
> &mdash; https://github.com/standard/semistandard

tsconfig.eslint.json
DerekNonGeneric: Conclusion: we can't use Esprima (the default parser used by ESLint) since it doesn't support top-level await yet.

```console
/mnt/c/orgs/tc39/tc39.github.io/_tasks/check-html.mjs
  4:20  error  Parsing error: Cannot use keyword 'await' outside an async function
```

This is will have to use [`@babel/eslint-parser`](https://www.npmjs.com/package/@babel/eslint-parser) then.

tsconfig.base.json
DerekNonGeneric: There is a slight discrepancy here that I would like to note with the config you proposed above. For the browser code, you are configuring it to run as though it is ES2021, but for compatability reasons that cannot be the case. This is going to need to use [`babel-preset-env`](https://babeljs.io/docs/en/babel-preset-env) + [Browserslist](https://babeljs.io/docs/en/babel-preset-env#browserslist-integration). Shouldn't be a problem, though.

.config/babel.mjs
```diff
@@ -0,0 +1,13 @@
+export default {
+  'plugins': ['@babel/plugin-syntax-top-level-await'],
+  'presets': [
+    [
+      '@babel/preset-env',
+      {
+        'targets': {
+          'esmodules': true
```
ljharb:
```suggestion
  plugins: ['@babel/plugin-syntax-top-level-await'],
  presets: [
    [
      '@babel/preset-env',
      {
        targets: {
          esmodules: true
```

.config/eslint.mjs
```diff
@@ -0,0 +1,103 @@
...
+  semi: true,
+  singleQuote: true,
+  tabWidth: 2,
+  trailingComma: 'es5',
+  useTabs: false,
+};
+
+const OFF = 0;
+const WARN = 1;
+const ON = 2;
```
ljharb: instead of these, eslint supports the strings "off", "warn", and "error". can we use these inline instead of making constants?

.eslintrc.js
```diff
@@ -0,0 +1,4 @@
+/** @module {CjsModule} */
```
ljharb: what is the purpose of this jsdoc comment? Every `.js` file in a node module is a CJS module.

.eslintrc.js
```diff
@@ -0,0 +1,4 @@
...
+'use strict';
+require = require('esm')(module);
```
ljharb: this is confusing to me. you're reassigning the modules' `require` function?

it seems like it'd be much simpler to put the config in this `.js` file directly rather than needing the `esm` indirection.

package.json
```diff
@@ -22,9 +23,19 @@
   },
   "homepage": "https://tc39.es/",
   "engines": {
-    "node": ">=14.8"
+    "node": ">=14.8.0"
```
ljharb: this is fine but also unnecessary; `>=14.8` and `>=14.8.0` are identical.

package.json
```diff
@@ -5,8 +5,9 @@
   "description": "Draft for TC39 website",
   "main": "index.js",
   "scripts": {
+    "format:js": "eslint --ext=.js,.mjs . --fix",
```
ljharb:
```suggestion
    "format:js": "npm run lint:js -- --fix",
```

.eslintrc.js
DerekNonGeneric: I originally started doing this because I wanted to omit file extensions from my config files. This was before the import assertions proposal made any progress. Still, I noticed how convenient it was to quickly identify the context of my JS module code w/o having to think about Node's resolution algorithms. I will get rid of it.

.config/eslint.mjs
DerekNonGeneric: We could, but this seems slightly more elegant to me. What is the advantage of using the strings? I can define the constants above the variables if that would be more technically correct. Is the goal to make the file less lines?

package.json
DerekNonGeneric: Known, but I used the exact version number here to help explain why there is so much code in here to support top-level await.

.eslintrc.js
DerekNonGeneric: Yeah, this is a little tough to grasp. This is colloquially called "hijacking require" and there are other ways of doing it.

This is preferable because now we get to use ES modules for our config. 😄

There is another way we can do this w/o depending on the `esm` package. Have you seen [my StackOver answer](https://stackoverflow.com/questions/48624588/use-export-default-in-eslintrc-js-file-instead-of-module-exports/57641788#57641788)?

.config/babel.mjs
DerekNonGeneric: Strange that this wasn't picked up by Prettier since we have defined `quoteProps: 'as-needed'` in the ESLint config file.

.eslintrc.js
DerekNonGeneric: This is not yet addressed. I know why this is not appreciated in this repo and I will fix it.

ljharb: Why does it matter that we’re using ESM for the config? What advantage does that offer? Static CJS and static ESM are conceptually identical.

.config/eslint.mjs
ljharb: The advantage is then it looks like an idiomatic eslint config, instead of having constants that are only defined in this one config.

.eslintrc.js
DerekNonGeneric: > Why does it matter that we’re using ESM for the config?

It matters because ESM is a standard.

> What advantage does that offer?

The advantage is that we will be using spec-compliant module syntax.

> Static CJS and static ESM are conceptually identical.

These wouldn't be static. JSON is a static data interchange format.

.config/eslint.mjs
DerekNonGeneric: A more developed website will eventually have a dynamic Babel config too. One for development, one for production.

.eslintrc.js
ljharb: CJS is a standard too, in node, which is the only place this config will ever be used.

Both are static when you use them statically, and both are dynamic when used dynamically (both support `import()`, for example). ESM offers no "purity" advantage here, just added complexity.

.config/eslint.mjs
ljharb: That's not my experience; you can (and one usually does) configure both in the same static Babel config pivoting on the NODE_ENV env var.

.config/babel.mjs
DerekNonGeneric: Oh, I just remembered why this happens! Because we need to tell ESLint not to ignore dotfiles.

.eslintrc.js
DerekNonGeneric: Okay, I will change it to use CommonJS. I thought this was leading to JSON configs, which loses comments.

.config/eslint.mjs
DerekNonGeneric: Yes, that's what I meant, actually. One config file, but separate configs per environment. I would consider that approach to be pretty dynamic, though. If that's not dynamic, I don't know what is. 😅

DerekNonGeneric: I guess the outstanding questions I have are whether the use of variables in the root `.eslintrc.js` file makes it _dynamic_, which detracts from its idiomaticness.

DerekNonGeneric:
@ljharb, you can disregard the second review &mdash; all of your concerns have been addressed.

This is pretty much as bare-bones as it gets. I don't think simplifying any further would be advantageos.

.eslintrc.js
```diff
@@ -0,0 +1,94 @@
+'use strict';
+
```
ljharb: not a big deal, but this now doesn't have to be JS, it can be a nice simple static `.eslintrc` :-)

_tasks/.babelrc.json
```diff
@@ -0,0 +1,15 @@
+{
+  "plugins": ["@babel/plugin-syntax-top-level-await"],
+  "presets": [
```
ljharb: i wonder if this could be made to `extends` the parent babelrc rather than replacing it?

ljharb (approved):
Thanks, I think you've resolved all my concerns, and I really appreciate your patience and persistence working through them.

Remaining comments would be nonblockers had I the place to block here; i'll leave it to the repo maintainers to review further.

.eslintrc.js
DerekNonGeneric: Alright, nice and simple is great &mdash; SGTM.

It should probably be noted that once rules are added in the future, it may be a bit difficult to understand w/o comments.

Also, I've heard that not having extensions is an anti-pattern. What do you think about `.eslintrc.json` instead?

_tasks/.babelrc.json
DerekNonGeneric: Yeah, that would be preferable. I'll look into it.

.eslintrc.js
ljharb: `.eslintrc` supports JSON5, so it can have comments just fine.

It's fine if you want to include the extension; i never include it in my projects, and it feels incorrect to me to have JSON5 in a `.json` file.

package.json
```diff
@@ -6,7 +6,7 @@
   "main": "index.js",
   "scripts": {
     "lint": "npm run lint:js && npm run lint:scss",
-    "lint:js": "prettier -c --loglevel warn **/*.js",
+    "lint:js": "eslint --ext=.js,.mjs .",
```
DerekNonGeneric, line 10: @codehag, I don't know if you want to lint the JSON or not, but if so, would you like it done in a separate package run script or should I add the extension here?

package.json
```diff
@@ -5,8 +5,9 @@
   "description": "Draft for TC39 website",
   "main": "index.js",
   "scripts": {
+    "format:js": "npm run lint:js -- --fix",
     "lint": "npm run lint:js && npm run lint:scss",
-    "lint:js": "prettier -c --loglevel warn **/*.js",
+    "lint:js": "eslint --ext=.js,.mjs .",
```
DerekNonGeneric, line 9:
```suggestion
    "lint:js": "eslint --ext=.js,.json,.mjs .",
```

Do I add the `.json` extension here? ESLint is already set up to lint JSON, but I'm unsure if this should be its own package run script or if this should be included in this line.

DerekNonGeneric: If that is not desirable, I will rebase to remove https://github.com/tc39/tc39.github.io/pull/213/commits/9110a57a4303c4437150e51b54920d6f03712663, let me know.

ljharb: Seems reasonable to me to include it, if it can be linted :-) altho any attempt to consume it would fail if invalid, so i'm not sure it's useful.

DerekNonGeneric: I was kinda hoping that this would notify about improper indentation (e.g., using tabs instead of spaces or not indenting at all), but it doesn't seem to be working configured this way. Any idea what I've done wrong?

DerekNonGeneric: Oh dang, I just read https://github.com/prettier/eslint-plugin-prettier/issues/294 and it doesn't seem possible to do it this way. :(

DerekNonGeneric:
Pretty sure this is what this PR is supposed to look like.
Might put these changes in a fresh PR, but until then&hellip; :)

thats-all-folks

DerekNonGeneric:
Please refer to the second take on this PR that is currently-open instead.

It reverts some of the early decisions we made:
- putting `.prettierrc` options in the `.eslintrc`
- attempting to style-check JSON using ESLint
- et al.

We actually do want to have a `.prettierrc` populated w/ this config for JSON style-checking.

---

Another thing that I did not address in this PR or the new one was creating a GitHub Action. I will have to followup w/ @codehag on that since I don't understand the reasoning behind it since Travis is already in use. Thanks so much for the tips though. I'm very pleased w/ the simplicity of the solution that resulted and welcome further scrutiny.

ljharb:
(linking to #214)

codehag:
>  I will have to followup w/ @codehag on that since I don't understand the reasoning behind it since Travis is already in use.

Travis is better since it is what we use -- I was saying actions because I have that on another project and forgot that we were using travis here.

Issue #223, opened 2020-12-03
[Bug] The page is empty


Jack-Works:
![image](https://user-images.githubusercontent.com/5390719/100956016-93e85780-3552-11eb-9626-fb584e404259.png)

codehag:
ah, that is unexpected.

codehag:
Temporary fix is online, full fix is incoming.

codehag:
We are back up and running:

tc39.es
tc39.es/de/
tc39.es/ru/
tc39.es/zh-Hans/

We need a redirect for zh-Hans to zh-Hans/ -- the other locals do this automatically but not this one.

PR #224, opened 2020-12-03
Fix multilang site deploy

codehag:
This addresses #223

The problem: 

In order to use plugins, we need a custom jekyll build instead of github pages builds. This is documented here: https://jekyllrb.com/docs/continuous-integration/github-actions/

And I 💃 forgot 💃 .

This is solved by using a github action for deployment to a custom branch (usually gh-pages). There are several such actions, I went with what was described in the jekyll documentation. One difference is -- rather than using a personal access token, I am using a github token. 

I also updated to readme so if anything goes wrong I get pinged. cc @Jack-Works for a second pair of eyes before merging.


codehag:
Unfortunately, there is no way to test this before deploying. I have a test branch here:
http://hag.codes/test/
http://hag.codes/test/de
http://hag.codes/test/ru
http://hag.codes/test/zh-Hans

The css is broken due to the way we created our urls -- this would be a separate fix that is unrelated to this one. It runs as expected locally.

codehag:
This change made me update all sorts of things, including netlify to the latest distribution.

chicoxyzzy (approved):

PR #229, opened 2020-12-04
proposed changes for German translations

Regaddi:
@codehag 

Here are some proposals to improve the readability in German on the website.
Translating technical jargon is a generally tough challenge, especially when mostly talking about that stuff in English 😅 

I'm happy to adjust the changes based on other peoples input ☺️ 

I'm not sure if changing the presented dates month names does any harm to any running automation, but I felt like displaying those in German, too, is slightly more accurate and also matches the example tags, that are rendered above the list.


codehag:
This looks really great to me! I think in the future we will just localize a date string, but for now this is great.

Regaddi:
Added a few more changes, since I had issues running the dev env locally. Seeing the changes in context is definitely valuable 😄

_data/de/site.yml
```diff
@@ -4,20 +4,20 @@ navigation:
     title: TC39
     agenda: Bevorstehende Agenda
     notes: Notizen der letzten Sitzung
-    sub-menu: Umschalte das Untermenü für TC39
+    sub-menu: TC39 Untermenü umschalten
```
Regaddi, line 7: It would probably sound better if we had separate texts for opening and closing a sub-menu, since `umschalten` is a very uncommon term.

`TC39 Untermenü öffnen` and `TC39 Untermenü schließen` would work much better and also give context about the current state of that submenu.

_data/de/site.yml
```diff
@@ -29,8 +29,8 @@ footer:
   archives: Archiv
   process: Prozess
   coc: Code of Conduct
-  feedback: Wir suchen nach Feedback
-  feedback-prompt: Bitte öffnen Sie <a href="https://github.com/tc39/tc39.github.io/issues/new" target="_blank">ein Issue </a> in <a href="https://github.com/tc39/tc39.github.io" target="_blank">unser Website-Repository</a>.
+  feedback: Wir sind auf der Suche nach Feedback.
```
Regaddi, line 32: There are certainly alternatives. I tried to keep it close to the original English phrasing.

codehag: I think, we can always use what is more common in the locale's language if it is more natural. But whatever you choose would work for me!

PR #237, opened 2021-01-02
account for unpublished notes

DerekNonGeneric:
Modified info crawler algorithm `presenceObject.url` to be optional to account for notes that have yet to be published (non-link emoji). Once the data structure is serialized to YAML, the field will not be present in the data structure.

### DIFF
--- a/_includes/proposals.html
+++ b/_includes/proposals.html
@@ -59,9 +59,13 @@
         <ul class="featurelist__item__status featurelist__item__tags">
           {% if proposal.presented | size %}
           {% for presentation in proposal.presented %}
-            <li class="featurelist__item__presented featurelist__item__tag">
-              <a href="{{ presentation.url }}" title='{{ site.data["proposals"]["last-presented-alt"] }}'>{{ presentation.date }}</a>
-            </li>
+          <li class="featurelist__item__presented featurelist__item__tag">
+            {%- if presentation.url -%}
+            <a href="{{ presentation.url }}" title='{{ site.data["proposals"]["last-presented-alt"] }}'>{{ presentation.date }}</a>
+            {%- else -%}
+            <span>{{ presentation.date }}</span>
+            {%- endif -%}
+          </li>
           {% endfor %}
           {% endif %}
           {% if proposal.tests | size %}
--- a/_sass/_featurelist.scss
+++ b/_sass/_featurelist.scss
@@ -15,11 +15,11 @@
     }
     &__tag {
       display: inline-block;
+      font-size: 0.8rem;
+
       a {
         color: white;
-        display: block;
-        font-size: 0.8rem;
-        margin-top: 0.5rem;
+        margin-top: 0.5rem;  
         opacity: 1;
         padding-left: 0.5rem;
         padding-right: 0.5rem;
@@ -32,6 +32,17 @@
         &:focus {
           background-color: transparent;
         }
+      }
+
+      span {
+        background-color: $color-dim-gray;
+        border: 2px solid $color-dim-gray;
+        border-radius: 10px;
+        color: white;
+        font-weight: 550;
+        padding-left: 0.5rem;
+        padding-right: 0.5rem;
+        margin-top: 0.5rem;
       }
     }
     &__status {


_includes/proposals.html
```diff
@@ -60,7 +60,7 @@ <h4 class="featurelist__item__title flex-grow">
           {% if proposal.presented | size %}
           {% for presentation in proposal.presented %}
             <li class="featurelist__item__presented featurelist__item__tag">
-              <a href="{{ presentation.url }}" title='{{ site.data["proposals"]["last-presented-alt"] }}'>{{ presentation.date }}</a>
+              <a {% if presentation.url %}href="{{ presentation.url }}" {% endif %}title='{{ site.data["proposals"]["last-presented-alt"] }}'>{{ presentation.date }}</a>
```
ljharb: Seems like if there’s no url, it shouldn’t be an a tag at all.

DerekNonGeneric: It seems to be perfectly valid (the HTML validator never complained). Apparently, the name is &ldquo;placeholder hyperlink&rdquo;.

Refs: https://stackoverflow.com/questions/5292343/is-an-anchor-tag-without-the-href-attribute-safe

DerekNonGeneric: If we were to change the tag to something else (like span), we would have to add the CSS for it too, which is not ideal.

ljharb: i don’t think anything should look like a link if it’s not going to navigate anywhere, so we’d need css changes regardless.

DerekNonGeneric: - [ ] [remove href](https://css-tricks.com/how-to-disable-links/#surefire-way-remove-the-href)
- [ ] [disabled](https://css-tricks.com/how-to-disable-links/#thats-not-enough-i-want-something-more-complex-so-that-i-can-look-smarter)
- [ ] [over-engineered](https://css-tricks.com/how-to-disable-links/#bring-in-the-javascript)
- [ ] something else

Someone wrote a blog post! 😂

DerekNonGeneric: Definitely agree, though. Disabled links should not look identical to non-links. +1 on disabled styled.

DerekNonGeneric: We could even remove the color from the link.

DerekNonGeneric:
Marking as a draft for now due to being incomplete (also currently low-priority due to there being no non-links).

_includes/proposals.html
DerekNonGeneric: That blog post is actually quite wrong&hellip; Deleted [the link](https://css-tricks.com/how-to-disable-links/#bring-in-the-javascript) and the choices it offered.

Notice that the [`not-allowed` value](https://developer.mozilla.org/es/docs/Web/CSS/cursor#Values) for `cursor` is under the &ldquo;drag & drop&rdquo; section. So, best not to do that (if we want to be technically correct).

P.S. I see non-link and disabled as the same thing.
The suggestion I propose is to make it grayscale.

ljharb: I’m not sure it should be too light - the date is the important part, the notes are just extra background info.

DerekNonGeneric: Thinking about something along the lines of the following (we already have the variable from before).

```scss
a:not([href]) {
    color: $color-dim-gray;
}
```

Refs: https://stackoverflow.com/questions/5128711/styling-a-without-href-attribute

ljharb: I’d still prefer not using an anchor at all if there’s nowhere to navigate - that it’s technically valid isn’t really interesting to me.

DerekNonGeneric: I will do as you wish (although am not certain on the reasoning).

ljharb: I would find it confusing to have a thing that looks and acts like a link, that has a right-click menu and hover behavior like a link, but that i can't open in a new tab, click to navigate, download the target, or save to bookmarks.

chicoxyzzy: Anchor without href is semantically weird. Also I think that it's not good for accessibility

DerekNonGeneric: Here's how it looks visually.

<br>

image

_sass/_featurelist.scss
```diff
@@ -33,6 +33,17 @@
           background-color: transparent;
         }
       }
+
+      span {
+        background-color: $color-dim-gray;
+        border: 2px solid $color-dim-gray;
```
ljharb, line 39: i'd use a darker color here; the date is important info, not just the notes.

ljharb (approved):

_sass/_featurelist.scss
DerekNonGeneric: There is a preview of how it looks in the commit message. Have you seen it?

ljharb: ah, no, i hadn't; i don't usually expand commit message descriptions. thanks, this looks great.

chicoxyzzy (approved):
LGTM 👍

PR #255, opened 2021-05-28
Add Japanese translation for stage 3, May 2021

smorimoto:


smorimoto:
@akihikodaki Could you please review this to help me?

_data/ja/stage3.yml
```diff
@@ -217,7 +217,7 @@
     - Tierney Cyren
   champions:
     - Tierney Cyren
-  description: Proposal for an Object.hasOwn() method to make Object.prototype.hasOwnProperty() more accessible.
+  description: よりObject.prototype.hasOwnProperty()へアクセスしやすくするためのObject.hasOwn()のプロポーザルです。
```
akihikodaki: I suggest to translate "アクセス" to ordinary Japanese words, even if it sounds slightly different. Translations can be difficult to read if you are not careful, so let's translate with "readability-first" (of course under the condition that it does not sound incorrect).
"へアクセス" may be replaced with "を利用".
I think "提案" is more understandable than "プロポーザル", but the word is used elsewhere so it should be used anyway.

_data/ja/stage3.yml
```diff
@@ -231,7 +231,7 @@
     - Shu-yu Guo
   champions:
     - Shu-yu Guo
-  description: ArrayBuffers have enabled in-memory handling of binary data and have enjoyed great success. This proposal extends the ArrayBuffer constructors to take an additional maximum length that allows in-place growth and shrinking of buffers. Similarly, SharedArrayBuffer is extended to take an additional maximum length that allows in-place growth. The transfer method is also re-introduced here as a standard way to detach ArrayBuffers, perform zero-copy moves, and to "fix" resizable ArrayBuffer instances to ArrayBuffer instances.
+  description: ArrayBuffersは、バイナリデータをインメモリで処理できるようにし、大きな成功を収めました。このプロポーザルは、ArrayBufferコンストラクタを拡張することで、バッファのインプレースの拡張と縮小を可能にする追加の最大長を取得可能にします。同様に、SharedArrayBufferは、インプレース拡張が可能な追加の最大長を取得できるよう拡張されています。Transferメソッドは、ArrayBuffersをデタッチし、ゼロコピーによる移動を行い、リサイズ可能なArrayBufferインスタンスをArrayBufferインスタンスとして"固定"するための標準的な方法としてもここで再導入されています。
```
akihikodaki: - "取得" does not sound a right translation for "take". It sounds like "get", but the functionality is more like a setter.
- "追加の最大長" is somewhat awkward. You cannot add yet another "maximum length of an array", can you? The criticism applies to the original text, but we can take advantage of Japanese grammar to fix the issue. Also, in my experience, "additional" is usually better to be translated to words else "追加" such as "他の", "別に", "新たに", etc.

So I suggest "バッファのインプレースの拡張と縮小が可能な最大長を新たに受け取れるようにします".

"transfer" is now capitalized, probably because it is now at the beginning of the sentence, but I prefer not to capitalize the word, as it represents `ArrayBuffer.prototype.transfer`. Japanese-language readers would not care even if the beginning of the sentence is capitalized anyway.

akihikodaki (changes requested):
The meaning of the translations is correct in general as far as I can tell. I have a few suggestions to make the translations more idiomatic Japanese.

smorimoto:
Thanks a lot! Everything makes perfect sense.

smorimoto:
@codehag I think we can merge this now.

_data/ja/stage3.yml
```diff
@@ -231,7 +231,7 @@
     - Shu-yu Guo
   champions:
     - Shu-yu Guo
-  description: ArrayBuffers have enabled in-memory handling of binary data and have enjoyed great success. This proposal extends the ArrayBuffer constructors to take an additional maximum length that allows in-place growth and shrinking of buffers. Similarly, SharedArrayBuffer is extended to take an additional maximum length that allows in-place growth. The transfer method is also re-introduced here as a standard way to detach ArrayBuffers, perform zero-copy moves, and to "fix" resizable ArrayBuffer instances to ArrayBuffer instances.
+  description: ArrayBuffersは、バイナリデータをインメモリで処理できるようにし、大きな成功を収めました。このプロポーザルは、ArrayBufferコンストラクタを拡張することで、バッファのインプレースの拡張と縮小が可能な最大長を新たに受け取れるようにします。同様に、SharedArrayBufferは、インプレース拡張が可能な追加の最大長を取得できるよう拡張されています。transferメソッドは、ArrayBuffersをデタッチし、ゼロコピーによる移動を行い、リサイズ可能なArrayBufferインスタンスをArrayBufferインスタンスとして"固定"するための標準的な方法としてもここで再導入されています。
```
rkirsling: `ArrayBuffers` → `ArrayBuffer` x2

rkirsling (changes requested):
Just one comment from me.

_data/ja/stage3.yml
smorimoto: Oops, indeed they should be fixed!

_data/ja/stage3.yml
akihikodaki: You should also do the same change for "同様に、SharedArrayBufferは、インプレース拡張が可能な追加の最大長を取得できるよう拡張されています。".

Although I have not noticed the last time, I think you should replace quotes (") around 固定 with 「」 as it's Japanese.

codehag:
Thanks, everyone -- ready to merge?

smorimoto:
I believe so

rkirsling (approved):

PR #256, opened 2021-05-28
Improved translations

codehag:
We have two translations in progress which will be merged into this -- please do not worry about rebasing! I will take care of that.

This patch should make translation a bit easier. Any shared information will fall through to the base stage3 yml file. I have also added a small readme about how the translations work. Feedback welcome @smorimoto @septs @chicoxyzzy @maramigin-e @akihikodaki 

@septs one note: I saw that for our chinese translations, we are using english names for the months. I left that as is -- I think this was a decision due to the months not mapping exactly? Let me know if that is correct or no.


akihikodaki (commented):
I'm wondering if all of `translation*.md` going to have reviewer lists. There would be (n * m) entries where n is the number of languages and m is the number of reviewers, which can be too big for humans to maintain in my opinion.

codehag:
> I'm wondering if all of `translation*.md` going to have reviewer lists. There would be (n * m) entries where n is the number of languages and m is the number of reviewers, which can be too big for humans to maintain in my opinion.

I am open to suggestions: should we have just one file?

akihikodaki:
> I am open to suggestions: should we have just one file?

Maybe we could have `translation/guidline_(replace with a BCP47).md` and `translation/reviewers.md` instead.

maramigin-e:
I think maybe create a file for each language with a specific guide on how best to translate and the number of contributors

PR #268, opened 2022-01-13
fix #265, replaced vnu-jar with html-validate

### LINKED ISSUE #265: replace vnu-jar with html-validate
https://www.npmjs.com/package/html-validate is a Node-based alternative to the vnu-jar HTML validator. The current setup requires setting up Java on a system, which is very unintuitive for contributors and is probably also slowing CI builds down.

This would simplify the build process, which would make it easier to contribute and move towards eventually simplifying the setup so we could remove Docker and GNU Make.

Moves us closer to #264.

osBins:
Uninstalled the npm package `vnu-jar` and installed `html-validate` as requested in issue #265. @ryzokuken let me know if any changes are to be made!

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,11 +34,11 @@
     "eslint-config-prettier": "6.14.0",
     "eslint-plugin-compat": "3.8.0",
     "eslint-plugin-prettier": "3.1.4",
+    "html-validate": "^6.1.6",
     "prettier": "^1.17.1",
     "stylelint": "^10.0.1",
     "stylelint-config-prettier": "^5.2.0",
     "stylelint-config-standard": "^18.3.0",
-    "stylelint-order": "^3.0.0",
-    "vnu-jar": "20.6.30"
+    "stylelint-order": "^3.0.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "lint:js": "eslint --ext=.js,.mjs .",
     "format:js": "npm run lint:js -- --fix",
     "lint:scss": "stylelint _sass",
-    "check:html": "bundle exec jekyll build && node _tasks/check-html.mjs",
+    "check:html": "html-validate _site/index.html",
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "repository": {
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "lint:js": "eslint --ext=.js,.mjs .",
     "format:js": "npm run lint:js -- --fix",
     "lint:scss": "stylelint _sass",
-    "check:html": "bundle exec jekyll build && html-validate _site/index.html",
+    "lint:html": "bundle exec jekyll build && html-validate _site/index.html",
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "repository": {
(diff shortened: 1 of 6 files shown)


jasonwilliams (approved):

jasonwilliams:
There will probably need to be a script to run the html validation against how it was ran before. So you can `npm run validate`

you would need to look at what the old one was doing

ryzokuken:
@osBins check out https://github.com/tc39/tc39.github.io/blob/master/_tasks/check-html.mjs. Since `html-validator` can be used from the CLI, you might as well just get rid of the file and replace the npm script by a single command.

osBins:
Alright. Can you tell me how the script is to be written in package.json? 
```
"check:html": "npm exec html-validate index.html",
```  
Is this correct or is there a particular way to write the path to index.html?

ryzokuken:
`"check:html": "html-validate _site/index.html",`

ryzokuken:
@osBins my bad, we still need to run jekyll. 

```
"check:html": "bundle exec jekyll build && html-validate ..."
```

ryzokuken (approved):
LGTM. Let's see how CI proceeds.

ryzokuken:
The new validation config is much stricter than the old one. Can we tweak the config to be functionally similar?

osBins:
Right, looking into it.

osBins:
I think I found it on [html-validate docs](https://html-validate.org/rules/presets.html). I need to specify a preset to configure the strictness of validation. Which one should we use?   
^Or let me know if I got it wrong.

ryzokuken:
@osBins please use https://html-validate.org/rules/presets.html#html-validate-standard

osBins:
Looks like the checks are still failing. Are we using the correct command for the script?  
https://html-validate.org/usage/index.html

ryzokuken:
Yes, it runs perfectly but finds more issues with our HTML code.

Check https://app.travis-ci.com/github/tc39/tc39.github.io/builds/244771032#L562-L571 for the warnings. Perhaps try manually setting these options off too? Beyond extending the "standard" preset?

osBins:
Had to rename a commit due to having missed out a word. The build has succeeded. Kindly check it out and let me know if anything else needs to be done.

ryzokuken (approved):
Great, let's move ahead! @jasonwilliams @smorimoto what do you folks think?

package.json
```diff
@@ -9,7 +9,7 @@
     "lint:js": "eslint --ext=.js,.mjs .",
     "format:js": "npm run lint:js -- --fix",
     "lint:scss": "stylelint _sass",
-    "check:html": "bundle exec jekyll build && node _tasks/check-html.mjs",
+    "check:html": "bundle exec jekyll build && html-validate _site/index.html",
```
jasonwilliams: would "lint:html" be more consistent with the above scripts? Does it have to be "check"?

ryzokuken: SGTM, let's rename it to `lint:html`. @osBins also remember to change in the CI script.

jasonwilliams:
If this issue replaces the java validator does that mean the java dependency can be removed here? Or are you planning to remove that at another time?

ryzokuken:
Right, good point @jasonwilliams. Let's all get rid of the Java dependency in this PR.

osBins:
Changed the script name to `lint:html` for consistency, as pointed out by @jasonwilliams.  
Not sure about how to get rid of the Java dependency as of now. Can you help me out here @ryzokuken?

jasonwilliams (approved):

ryzokuken (approved):
LGTM.

PR #269, opened 2022-01-22
Eleventy migration

### LINKED ISSUE #264: tc39.es v2
As discussed in Matrix, we have a plan to rework the website. The current goals are still unclear and not too broad, but here are a few:

- Easy to contribute: The project is too chaotic. It must be able to bootstrap just with `npm i`.
- Responsive and accessible: As the title implies, there should be a better experience (particularly on mobile devices).
- Integrated with the dataset (optional): We already have the dataset and it would be nice if we could remove the need for manual updates based on it (except for translations), but we don't have any specific ideas on this yet, so it's completely optional.

### LINKED ISSUE #266: Do we need a HTML Linter?
Follow up from https://github.com/tc39/tc39.github.io/issues/265 

Is linting the HTML a requirement for this project? It's not something I've often seen employed in other projects similar to this.
It feels overkill but maybe there was a reason behind it.

Can the HTML changes be checked through code review? 
Are there tools that are generating the raw HTML anyway?

Related to: https://github.com/tc39/tc39.github.io/issues/264

### LINKED ISSUE #267: Migrate from Travis to Github Actions
Github Actions has much better transparency (just click Actions at the top), so would be much better for contributors of this project. 

The aim would be to transform this config https://github.com/tc39/tc39.github.io/blob/master/.travis.yml into a `.github/workflows/release` action

It might be worth doing this after migrating from Ruby, so there's only Node to setup.

- https://github.com/actions/setup-node#usage

Related to: https://github.com/tc39/tc39.github.io/issues/264

jasonwilliams:
https://www.11ty.dev/

eleventy is pretty close to jeykll so I belive anyone contributing before won't feel lost. Most config/structure is the same. The only major difference is internationalization which I believe may mean us moving some things around but that's about it. The templating system is called liquid and seems to be almost identical to Jeykll's templating system only small changes in the tag names. The data is in JSON rather than yml but this is minor.

There's a lot of file changes in this diff but a majority of them are converting the yml files to json.
## Outstanding

- [x] Octicons need to be added back in (unless we don't care about these being gone)
- [x] Replace travis with Github Actions
- [x] Adding back in l10n (in progress) 
- [x] Check HTML output
- [x] clean up

## To Test
- Checkout
- $ npm ci
- $ npm start

Fixes https://github.com/tc39/tc39.github.io/issues/267
Fixes https://github.com/tc39/tc39.github.io/issues/266
Fixes https://github.com/tc39/tc39.github.io/issues/264


jasonwilliams:
deploy/netlify  would need to be modified but i think only @codehag can do that

PR #279, opened 2022-04-03
Update deps

smorimoto:


codehag (approved):
I didn't do an in-depth check, but it renders and skimming it, this looks fine.

jasonwilliams:
There's a lot of formatting changes here, is it my understanding that prettier updated and all the rules have changed? Or is there something else at play?

jasonwilliams (approved):

smorimoto:
The overall diff is quite large, but the first two commits are almost all of the changes.

PR #280, opened 2022-04-04
Clean up the root directory by remove unused files

smorimoto:


smorimoto:
At least, as far as the Makefile is really concerned, we probably moved to the JS based system to get rid of it. So! We can and/or should remove it! The editorconfig is arguable, but it's not necessary at this very moment because it's easy to mess things up in relationship to Prettier and others. We can re-visit if we really need it.

ljharb:
why delete editorconfig instead of fixing the name to `.editorconfig`?

codehag (approved):
iiuc we no longer use docker, so this is fine.

codehag:
> why delete editorconfig instead of fixing the name to `.editorconfig`?

I don't think anyone is using it. I am not, but we can keep it if necessary.

smorimoto:
For example, if you change the eol or space settings in the Prettier configuration, which one takes precedence, editorconfig or Prettier? We don't always need more than we need unless we really need it.

ljharb:
editors, and github.com itself, read editorconfig, and don't generally read prettier settings.

I agree they should be kept in sync, but it's better to have two files not in sync, than to lack the universally recognized editorconfig.

smorimoto:
The editorconfig extension in some editors overwrites editor behaviour, and if it behaves differently to Prettier's configuration, things are quite tricky, and while I don't think there are many situations in this project that need editorconfig. So I don't think it will benefit from having that extra layer of complexity.

DerekNonGeneric (approved):
The preferences specified in the `.editorConfig` file never seem to be applied due to the file name being wrong, but even with that corrected, it would not be enforced by any tooling (at the moment). Seems fine to let it go for now until that gets implemented. I have been using [`editorconfig-checker`](https://www.npmjs.com/package/editorconfig-checker) for enforcement in my other projects, but that can probably come later if desired (it seems so).

The `Makefile`, on the other hand, appears to have become obsoleted with migration to Eleventy.

DerekNonGeneric:
I would like to circle back to this real quick just to make sure we have the proper resolution.

Regarding how Prettier interacts with options specified in `.editorconfig` files, the following is according to their official docs.

> If `options.editorconfig` is true and an [`.editorconfig` file](https://editorconfig.org/) is in your project, Prettier will parse it and convert its properties to the corresponding Prettier configuration. This configuration will be overridden by `.prettierrc`, etc.
>
> &mdash; https://prettier.io/docs/en/configuration.html#editorconfig

So, evidently, Prettier _does_ respect the options specified in `.editorconfig` files if properly configured, so I think it would be a good idea if we were to re-add it and do that here. Additionally, it would be ideal if we had the proper tooling set up to be sure that those options were enforced by CI independent of Prettier. The reason is that Prettier is only active on a limited subset of file formats while EditorConfig is able to handle all text files. I linked to that bit of tooling in my previous post.

What do you all think? If there are no objections, I might go for it if that's alright.

smorimoto:
I totally missed it! I would love to take a care of it when it gets open.

DerekNonGeneric:
> I totally missed it! I would love to take a care of it when it gets open.

@smorimoto, I got it started over at https://github.com/tc39/tc39.github.io/pull/293 &mdash; seems other open PR is currently handling re-adding of the `.editorconfig`.

PR #284, opened 2022-05-06
Add default favicon ico link

nairihar:
For my browser, the favicon isn't visible.
We need to include the default ico format too.

Screen Shot 2022-05-07 at 2 11 21 AM

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -1,5 +1,6 @@
 module.exports = function (eleventyConfig) {
   eleventyConfig.addPassthroughCopy('assets');
+  eleventyConfig.addPassthroughCopy('favicon*');
   return {
     dir: {
       input: './', // Equivalent to Jekyll's source property
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -8,6 +8,7 @@
     />
     <meta name="theme-color" content="#157878" />
     <title>TC39 – {{ site["title"] }}</title>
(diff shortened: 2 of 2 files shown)


ljharb (approved):
this is fine, but what browser is that that doesn't understand PNG favicons??

smorimoto:
@ljharb @nairihar The underlying issue is different in the first place and I've added a commit to address it.

smorimoto (approved):

smorimoto:
Thank you @nairihar.

DerekNonGeneric:
We seem to not currently be following the guidelines to be eligible for a favicon to be displayed next to search results.

> Your favicon must be a multiple of 48px square, for example: 48x48px, 96x96px, 144x144px and so on. SVG files don't have a specific size. Any [valid favicon format](https://en.wikipedia.org/wiki/Favicon#File_format_support) is supported. Google will rescale your image to 16x16px for use in search results, so make sure that it looks good at that resolution.
>
> **Caution**: Don't provide a 16x16px favicon.
>
> &mdash; https://developers.google.com/search/docs/advanced/appearance/favicon-in-search

It may not be clear from the GitHub web view, but the current [`favicon.ico`](https://github.com/tc39/tc39.github.io/blob/HEAD/favicon.ico) being served appears to be a multiple of a 16px square as well (containing 16x16px and 32x32px images) from what I was able to observe using IrfanView.

Issue #221, opened 2020-12-02
Structured pages (SEO)


septs:
> Google Search works hard to understand the content of a page. You can help us by providing explicit clues about the meaning of a page to Google by including structured data on the page. Structured data is a standardized format for providing information about a page and classifying the page content; for example, on a recipe page, what are the ingredients, the cooking time and temperature, the calories, and so on.

from https://developers.google.com/search/docs/guides/intro-structured-data

Use JSON-LD structured page for Google Search results?

Reference:
- https://schema.org/WebPage
- https://schema.org/ItemPage
- https://schema.org/Event
- more, see https://schema.org/docs/schemas.html

septs:
CC @codehag

codehag:
This sounds like a great idea. I don't have much experience in this area so I am happy to let you take the lead on this. Do you need any information or access to do that?

codehag:
We can likely have this localized as well

PR #287, opened 2022-07-22
add links to all alternate language pages in head

DerekNonGeneric:
List all current alternate language variants of the page (including
itself) in the `<head>` section, to aid search engines in directing
users to the most appropriate version by language or region.

Refs: https://developers.google.com/search/docs/advanced/crawling/localized-versions#html

---

This is not really a fix for #221, but another similarly-related SEO task I discovered had been missing while looking into the structured data recognized by search engines. I could have gone with the alternative way of [specifying this in a sitemap](https://developers.google.com/search/docs/advanced/crawling/localized-versions#sitemap), but am not sure whether one currently exists for these pages or whether doing it that way would make a noticeable difference here.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -8,6 +8,11 @@
     />
     <meta name="theme-color" content="#157878" />
     <title>TC39 – {{ site["title"] }}</title>
+    <link rel="alternate" hreflang="de" href="https://tc39.es/de/" />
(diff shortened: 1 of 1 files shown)


smorimoto (approved):

smorimoto:
I'm sure we can go with this, so I just merged, but please let me know if there are any problems.

PR #292, opened 2022-07-29
Power proposals from Dataset API fixes #290

### LINKED ISSUE #290: Populate proposal data from tc39/dataset
Populate the website with data from https://github.com/tc39/dataset

Eleventy Fetch seems useful (and for local development):
https://www.11ty.dev/docs/quicktips/cache-api-requests/

**Task:**
A new JS file would need to be added to https://github.com/tc39/tc39.github.io/tree/master/_data which has a similar format to the link above. It fetches proposal data and returns a format we're expecting. At build time this API would be fetched and the data can be iterated through in the template.

**Triggering Action from dataset**
Either https://github.community/t/triggering-by-other-repository/16163/2 or a cron action to rebuild the website at a set interval

**Issues:**
- Translations? I'm not sure we would have a solution for this if pulling from the API. I'm open to any ideas

**Related:**
- https://github.com/tc39/tc39.github.io/issues/199
- https://github.com/tc39/tc39.github.io/issues/60

jasonwilliams:
- Adds back editorConfig, its useful when committing from Windows to make sure LF is set, plus other settings with spaces/tabs.
- eleventyApi uses a .cache file to store the response ignore this.
- update dependencies

We could in theory add all the proposals to the website now but ill leave that out of scope for this.

Issue raised with dataset for missing data: https://github.com/tc39/dataset/issues/15

fixes #290


_includes/proposals.html
```diff
@@ -19,9 +19,7 @@ <h3>{{ proposals["title"] }}</h3>
         <li>
           {{ proposals["how-to-read"]["tags"]["last-presented"] }}:
           <div class="featurelist__item__presented featurelist__item__tag">
-            <a
-              href="{{ presentation.url }}"
```
smorimoto, line 23: No `href` is needed here.

_includes/proposals.html
```diff
@@ -48,27 +46,16 @@ <h3>{{ proposals["title"] }}</h3>
...
-{% for proposal in stage3 %}
+{% for proposal in stages %}
 <article class="featurelist__item">
   <div class="featurelist__item__intro">
     <header class="g-row">
       <h4 class="featurelist__item__title flex-grow">
-        <a href="https://github.com/tc39/{{ proposal[0] }}"
-          >{{ proposal[1].title }}</a
-        >
+        <a href="{{ proposal.url }}">{{ proposal.name }}</a>
```
smorimoto, line 54: `proposal.url` already has a complete URL.

_includes/proposals.html
jasonwilliams: Looks like you've removed this anyway?

smorimoto (approved):
I'm sure I have fixed all of them, so I have no objection.

.editorConfig
```diff
@@ -0,0 +1,30 @@
+# This file is for unifying the coding style for different editors and IDEs
```
DerekNonGeneric: Please be sure that we get the file name correct this time.

`.editorConfig` -> `.editorconfig`

Also, seems we are having excess unused config options and the Ruby and Makefile stuff could perhaps be trimmed.

smorimoto: OK. Done.

PR #293, opened 2022-07-30
ensure Prettier accommodates EditorConfig options

DerekNonGeneric:
Refs: https://prettier.io/docs/en/configuration.html#editorconfig

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,6 @@
 {
   "arrowParens": "always",
+  "editorconfig": true,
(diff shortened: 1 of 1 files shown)


DerekNonGeneric:
We may still need to add the npm run script for this, which seems to simply be as follows.

```json
"scripts": {
  "lint:editorconfig": "npx editorconfig-checker"
}
```

smorimoto (approved):

PR #294, opened 2022-07-30
Revert "ensure Prettier accommodates EditorConfig options"

smorimoto:
Reverts tc39/tc39.github.io#293


smorimoto:
We didn't really need this option: https://github.com/prettier/prettier/issues/6176#issuecomment-498808239

DerekNonGeneric:
Good catch, the docs are quite misleading in that regard.

PR #295, opened 2022-08-18
fix(stylelint): re-order presets correctly

DerekNonGeneric:
`stylelint-config-prettier-scss` disables all rules that conflict
with Prettier. Leave this preset at the bottom so that it can't be
overridden.

> Make sure to put it **last**, so it will override other configs.

Refs: https://github.com/prettier/stylelint-config-prettier-scss#installation

/cc @smorimoto

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -1,7 +1,7 @@
 {
   "extends": [
     "stylelint-config-standard-scss",
-    "stylelint-config-prettier-scss",
-    "stylelint-config-recess-order"
+    "stylelint-config-recess-order",
+    "stylelint-config-prettier-scss"
   ]
 }


DerekNonGeneric:
I am unsure if this change introduces anything actionable regarding re-styling, but CI is green, so it seems good to go assuming it is currently set up to lint on PR.

/cc @smorimoto

smorimoto (approved):

smorimoto:
Good catch!

smorimoto:
It would also be nice to add a format check step to CI...

DerekNonGeneric:
> It would also be nice to add a format check step to CI...

Yeah, how do you want to handle that? There are a few ways that come to mind&hellip;

Want to use a GitHub action to determine which formatter to run depending on the file extension or something?

PR #310, opened 2023-03-06
Correct outdated fields in `package.json`

smorimoto:
This patch corrects all outdated fields in `package.json` based on the npm
official specification.

Signed-off-by: Sora Morimoto <sora@morimoto.io>

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,9 +1,10 @@
 {
-  "name": "tc39-web-draft",
-  "private": true,
-  "version": "0.0.0",
-  "description": "Draft for TC39 website",
-  "main": "index.js",
+  "name": "@tc39/tc39.github.io",
+  "description": "Get involved in specifying JavaScript",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/tc39/tc39.github.io.git"
+  },
   "scripts": {
     "lint": "npm run lint:js && npm run lint:scss && npm run lint:format",
     "lint:js": "eslint --ext=.js,.mjs .",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,9 +1,20 @@
 {
-  "name": "tc39-web-draft",
-  "private": true,
+  "name": "@tc39/tc39.github.io",
```
ljharb, line 2: since private is true, i'm not sure why the name matters, but this is fine

package.json
```diff
@@ -1,9 +1,20 @@
...
   "version": "0.0.0",
-  "description": "Draft for TC39 website",
-  "main": "index.js",
+  "description": "Get involved in specifying JavaScript",
+  "homepage": "https://github.com/tc39/tc39.github.io",
+  "bugs": {
+    "url": "https://github.com/tc39/tc39.github.io/issues"
+  },
```
ljharb: all of these fields are only used when a package is published, so there's no need for them to be here

package.json
smorimoto: It was a kind of normalisation.

package.json
smorimoto: Indeed. I thought about getting rid of them, but I just made a pure correction here, somehow. On an extreme level, we can remove all the extra metafield except private.

jasonwilliams: Yeah I’m not sure if all of these are needed.

I would change or remove `maintainers` as I believe ujjwal, myself and potentially others are still around too. Yulia also created this site with myself, Ujjwal and some others initially so having the concept of a single “author” doesn’t make much sense in this context.

node we can keep for local development, license is fine, description and name are ok to keep.

smorimoto: Those opinions certainly sound reasonable. I reflected it in the latest commit.

package.json
```diff
@@ -20,19 +17,6 @@
     "prebuild": "npm run sass",
     "start": "npx @11ty/eleventy --serve --watch"
   },
-  "repository": {
```
ctcpip, line 23: suggest keeping `repository`

smorimoto: Given that this is an application, not a package, this field also seems unnecessary. What do you think?

ljharb: the repository field still works in applications with `npm repo`, so it's best to keep it.

smorimoto: OK - I just restored it

ctcpip (approved):
ambivalent about `homepage` being nixed, but LGTM

PR #312, opened 2023-03-06
✨ update agenda and notes links

### LINKED ISSUE #311: outdated links to agenda and meeting notes
under the first menu: tc39

are these updated manually?

ctcpip:
resolves #311

### DIFF
--- a/_layouts/menu.html
+++ b/_layouts/menu.html
@@ -16,7 +16,7 @@
       <li class="submenu-item">
         <a
           class="menu-link"
-          href="https://github.com/tc39/agendas/blob/main/2023/03.md"
+          href="https://github.com/tc39/agendas/blob/HEAD/2023/03.md"
         >
           {{ site["navigation"]["TC39"]["agenda"] }}
         </a>
@@ -24,7 +24,7 @@
       <li class="submenu-item">
         <a
           class="menu-link"
-          href="https://github.com/tc39/notes/tree/main/meetings/2023-01"
+          href="https://github.com/tc39/notes/tree/HEAD/meetings/2023-01"
         >
           {{ site["navigation"]["TC39"]["notes"] }}
         </a>
--- a/_layouts/menu.html
+++ b/_layouts/menu.html
@@ -16,7 +16,7 @@
       <li class="submenu-item">
         <a
           class="menu-link"
-          href="https://github.com/tc39/agendas/blob/HEAD/2022/09.md"
+          href="https://github.com/tc39/agendas/blob/main/2023/03.md"
         >
           {{ site["navigation"]["TC39"]["agenda"] }}
         </a>
@@ -24,7 +24,7 @@
       <li class="submenu-item">
         <a
           class="menu-link"
-          href="https://github.com/tc39/notes/blob/HEAD/meetings/2022-06"
+          href="https://github.com/tc39/notes/tree/main/meetings/2023-01"
         >
           {{ site["navigation"]["TC39"]["notes"] }}
         </a>


_layouts/menu.html
```diff
@@ -16,15 +16,15 @@
       <li class="submenu-item">
         <a
           class="menu-link"
-          href="https://github.com/tc39/agendas/blob/HEAD/2022/09.md"
+          href="https://github.com/tc39/agendas/blob/main/2023/03.md"
```
ljharb:
```suggestion
          href="https://github.com/tc39/agendas/blob/HEAD/2023/03.md"
```

_layouts/menu.html
```diff
@@ -16,15 +16,15 @@
...
         >
           {{ site["navigation"]["TC39"]["agenda"] }}
         </a>
       </li>
       <li class="submenu-item">
         <a
           class="menu-link"
-          href="https://github.com/tc39/notes/blob/HEAD/meetings/2022-06"
+          href="https://github.com/tc39/notes/tree/main/meetings/2023-01"
```
ljharb:
```suggestion
          href="https://github.com/tc39/notes/tree/HEAD/meetings/2023-01"
```
why is this one "tree" and the other one "blob" above?

ctcpip: `blob` is a file

_layouts/menu.html
ctcpip: I prefer the static ref over the symbolic

_layouts/menu.html
ljharb: ah, so the previous link here was wrong?

_layouts/menu.html
ljharb: The intention is to use links that don't hardcode the name of the default branch, to avoid the same churn that happened with master → main. We do the same thing org-wide, generally.

_layouts/menu.html
ctcpip: correct

_layouts/menu.html
ctcpip: `HEAD` is not actually documented anywhere and could break, although unlikely.

one subtle difference is that HEAD actually points to a commit id (the HEAD of course), and if you navigate around, all the links become commit-tree specific which is really annoying.  like if from there, I click the month, I go to `https://github.com/tc39/agendas/tree/9221e949320f570bc86405ec3a90310b219fb0bd/2023` instead of `https://github.com/tc39/agendas/tree/main/2023`

ljharb: I'm skeptical it's not documented anywhere, but either way all of TC39 relies on it already so I don't think that's a significant concern :-)

Re the commit ID, that's true, but why is that annoying?

ctcpip: well -- for one example where it was annoying was updating these links.  I went to the existing links, and clicked to navigate the repo to go to the most recent, and of course I did not get the links I needed.  I got:

`https://github.com/tc39/agendas/blob/9221e949320f570bc86405ec3a90310b219fb0bd/2023/03.md`
and
`https://github.com/tc39/notes/tree/c7fa95d3f2d0af68eaba9d388d5f79f2e166a4cc/meetings/2023-01`

ljharb: Fair. I'm not sure that's worth hardcoding breakable links tho for the common use case of "reading the notes"

ctcpip: they're not really breakable any more than the HEAD links are... even `master` links still redirect to `main`.  (there are some in this repo that should get changed as well)

ljharb: Those redirects will break if anyone ever accidentally creates a master branch, altho the fix is easy (delete that branch) so maybe now that github has implemented this redirection logic (which didn't exist for awhile after people were switching default branch names en masse) it's not that big a deal 🤔

ctcpip: Screenshot 2023-03-06 at 1 35 58 PM

currently nearly a 50/50 split betwixt these variations in code refs anyway

not that it means anything

🤷

ljharb: i think that's mostly in proposal repos; in the long-term archival ones, like proposals and agendas and the spec, we're pretty good about sticking to HEAD i hope

ctcpip: I'm not gonna die on this hill, so I'll change it but

ctcpip: some of those refs should be pointing directly to commit IDs, I imagine, but... trade offs...

in any case, I will change to HEAD per momentum in this usage here, but for the record, suggest we use `main` in general

ljharb: I'll think more on it, thanks

ljharb (approved):

PR #314, opened 2023-03-17
🇫🇷  - translation

ctcpip:
🇫🇷 translation provided by @AugustinMauroy
review and corrections provided by @ThomasDq


ThomasDq (approved):

smorimoto:
I'm not very fluent in French, so I can't do a strict "review" here, but at least it's definitely better than nothing!

smorimoto (approved):

PR #322, opened 2023-04-18
Update `.nvmrc` to use Node.js 20

smorimoto:


ctcpip:
I think this is fine, but why not stick with the active LTS for this?

smorimoto:
I don't have a very strong opinion, but since it's just a runtime for development in a way here, I thought there was no reason to stick to anything too much 🙂

ljharb:
If the intention is to stick to LTS, it shouldn't have `18`, it should have `lts/*` or `lts/hydrogen` :-)

ctcpip:
more of an aside than anything else, but for _reasons_, I always prefer to specify the exact version -- `x.y.z`

but, I am fine with any of the above for this.  will approve as is, but will approve any of these alternatives as well

ctcpip (approved):

smorimoto:
There is no guarantee that CI will always work well with LTS, so I would just specify the major version here 😑 (As you know, the JS ecosystem grows too fast!)

PR #337, opened 2023-07-12
Update npm deps

smorimoto:
Since there was a major update to Prettier, there is a lot of re-formatting around the HTML.


ctcpip:
can we just delete the notes folder?  appears to be dead code

ctcpip:
I updated the lint commands to ignore the notes folder, as it's for legacy url preservation anyway

ctcpip (approved):

smorimoto:
Makes sense!

PR #339, opened 2023-08-08
Twitter is X now

smorimoto:


smorimoto:
This seems a bit strange, but at least it's not our thing 🤷‍♂️

image

ctcpip:
I think we should not change this right now.  There is a good chance it will re-re-brand back to Twitter.  Also, the vast majority of places I still see the bird logo and twitter reference.  And it's still `twitter.com`

smorimoto:
Well, I have no objection to keeping this 🙂

_layouts/page_footer.html
```diff
@@ -83,23 +81,22 @@ <h3 id="connect-label">{{ site["footer"]["connect"] }}</h3>
...
-                ><svg
-                  height="24"
-                  aria-label="Twitter"
-                  class="octicon octicon-mark-twitter"
-                  viewBox="0 0 16 16"
+            <a href="http://x.com/tc39" target="_blank">
+              <span class="icon">
+                <svg
+                  width="24"
+                  aria-label="X formerly known as Twitter"
```
smorimoto, line 88: Of course it's well-taken into account.

ljharb:
We shouldn’t make this change ever; if we don’t say “Twitter” we should just remove the irrelevant reference.

DerekNonGeneric (approved):
This seems good to me. I have been following the re-brand closely and it is progressing well. Even my bills for Twitter Blue are changing to X.

ctcpip:
closing for now.  we can reopen if we want to proceed with this change in the future

smorimoto:
Now I feel this topic should be revisited. Many sites have updated it and it is no longer too early.

ljharb:
It’s not about being early; changing it to X is a political statement we shouldn’t make, ever.

smorimoto:
Hmm? Wait, does it have any political intent?

ljharb:
Using the name “X” implies support of Elon and his agenda. It should either say “twitter” or be removed entirely.

ryzokuken:
I don't think using the name Twitter while continuing to use the service regularly makes much of a difference at this point. I feel that we should switch to X on the website but if you're against us being on the platform altogether, I'd be supportive of that.

ljharb:
I don’t care about being on the platform; i only care that if it’s listed, it’s not referred to by “X”.

robpalme:
At this point it is a matter of fact that the service renaming has happened, so updating the website to be accurate on URLs and names is the appropriate thing to do.

If you believe the association between the service and the committee is a problem, that should be raised as a separate issue and we can discuss it. Personally I'm not aware of folk who would be sentimental about closing the rarely-used `@TC39` X/Twitter account, so feel free to start a Reflector thread proposing this.

ctcpip:
Continuing to defiantly refer to it as twitter is more of a political statement than referring to it by the rebranded name.  Independent non profit news agencies such as AP and AFP refer to it as X, and we should follow suit.

ljharb:
I don’t think the account should be closed at all; I’m not sure why that keeps being suggested.

I think that if it can’t be left as “Twitter” then it should be removed from the website, is all.

@ctcpip im not saying we can’t make any political statements; I’m saying that the statement made by calling it Twitter is aligned with our values whereas the statement made by calling it X is not.

ctcpip:
We are not making a statement by referring to it as X. We _are_ making a statement by referring to it as twitter.

We should change it to X, or alternatively perhaps refer to it as both, as news orgs did for some time, before ultimately only referring to it as X.

ljharb:
> We are not making a statement by referring to it as X

Clearly i strongly disagree.

> We are making a statement by referring to it as twitter.

Yes, our stated values require us to do so imo.

robpalme:
>perhaps refer to it as both

Thanks for suggesting this compromise, @ctcpip - I don't want this to get missed. So I guess we would call it _"X, formerly known as Twitter"_.  This seems the most reasonable resolution that meets the accuracy requirement.

@ljharb does this resolve your concern?

ljharb:
It’s the least bad alternative to leaving it as is or removing the reference altogether, but it’s still not ideal.

I continue to fail to see the motivation to making a change tho - who is that benefiting?

jasonwilliams:
> I continue to fail to see the motivation to making a change tho 

Because thats the name of the social network, they’re not going to name it back and our icon is outdated, so Sora offered to change it.

> Using the name “X” implies support of Elon and his agenda. 

I don’t agree with that, and I don’t see any organisation referring to it as “X” supporting any agenda or Elon personally. W3C have updated their icon on https://www.w3.org/, ISO the same https://www.iso.org/home.html ECMA also https://ecma-international.org/ these are not standards organisations supporting an agenda (last I checked). 

> @ctcpip im not saying we can’t make any political statements; I’m saying that the statement made by calling it Twitter is aligned with our values whereas the statement made by calling it X is not.

I don’t think either of those actions are “making a statement”. The fact this very PR exists shows that not everyone sees updating this icon being politically sensitive or sending a message. It’s outdated and needs changing.

We should re-open this PR.

ljharb:
@jasonwilliams there's lots of things that a subset of people find problematic that are thus problematic; some people not sharing the concerns of others does not invalidate them, or else our entire inclusivity and CoC efforts wouldn't exist.

smorimoto:
In my opinion, as a person who knows nothing about political stuff, I don't see what's wrong with just basing the display on facts. Linking everything to politics will only make us feel like a wild-goose chase 🤷‍♂️

PR #354, opened 2023-11-03
Adopt corepack

smorimoto:


ljharb (changes requested):
Why? npm is what ships with node, corepack (which is still experimental) isn't needed for it, and we absolutely don't want to pin to a specific version of npm ever.

smorimoto:
Using corepack and forcing developers to have consistent npm versions between them is not a bad choice in my experience. I did consider that it might not be ideal to specify a single version, but I think it is enough to just run `corepack up` to bump up the lower bound as needed.

smorimoto:
(will be) superseded by #356.

PR #356, opened 2023-11-10
Bump `engines.node` to `>=20`

smorimoto:
This change minimises my concerns, which I tried to address in #354, in a different way as reasonably and acceptably as possible.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
         "stylelint-config-standard-scss": "11.1.0"
       },
       "engines": {
-        "node": ">=18"
+        "node": ">=20"
       }
     },
     "node_modules/@11ty/dependency-tree": {
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "stylelint-config-standard-scss": "11.1.0"
   },
   "engines": {
-    "node": ">=18"
+    "node": ">=20"
   },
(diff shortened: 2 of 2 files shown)


smorimoto:
@ljharb This change does not strictly enforce the npm version, but the narrow lower bound of Node.js restricts the use of a wide range of bundled versions of npm in most scenarios.

jasonwilliams (approved):
This change is much more preferential to introducing corepack. @smorimoto can I ask the context behind this? Did something break compatibility somewhere?

smorimoto:
@jasonwilliams Some people have experienced lockfile versions being rolled back to the old one due to the old npm version. (and I received some PMs about that 🤷‍♂️)

ljharb:
Totally agree.

You can also set engines.npm if strictly necessary.

ljharb:
and, you can specify lockfileVersion in npmrc.

smorimoto:
> and, you can specify lockfileVersion in npmrc.

Oh, I didn't know about that. I don't really use npm these days, but it seems to be okay to use it recently.

smorimoto:
@ljharb Thank you for the great review to get things moving in a right direction!

ljharb:
It remains as it's always been, the best choice :-)

PR #388, opened 2024-01-31
Allow favicons to be loaded on non-root paths

smorimoto:
| Before | After |
| :-: | :-: |
| image | image |


ctcpip (approved):
Thanks! looks good; we should add more sizes in future

PR #274, opened 2022-02-16
Update links

moondef:
Update links to future meeting agenda and latest meeting notes

PR #396, opened 2024-02-17
Replace branch name `master` with `main` in links and fix translation docs

shuuji3:
Hello, this PR fixes two small issues:

1. Fix the link to a file in the `master` branch. GitHub web interface can redirect from `master` to `main` automatically but some user-agent fails (for example, GitHub Android app). This change ensures that users navigate to the correct page.

2. Delete the instruction to change `_config.yml` from `translation*.md`, which was removed by Eleventy migration and no longer exists (#269).


smorimoto:
This PR reminds me of the changes in #274 and #312. Maybe we should use `HEAD` instead of `main` to make it more robust to future changes.

@ljharb @ctcpip I believe you have any thoughts on this. What do you think?

ctcpip:
there are differences between them and while I prefer `main` for a number of reasons for most uses, like the fact that it actually results in links pointing to the tip of `main` rather than `HEAD` which will point to a specific commit id, the pattern of using `HEAD` has been established and it's not a hill I want to die on just yet 😄

ljharb:
I prefer HEAD since it will work regardless of the default branch name; the usability issue @ctcpip mentions exists, and either change is better than the wrong branch name.

smorimoto:
Fair enough! I want to unify here whether to use HEAD or branch names. I don't think changes like main becoming the de facto standard will happen often in the future, and I just favor branch names for now. If you all have no major objections to this idea, I will make a change to switch from HEAD to branch names for remaining parts in a subsequent PR.

smorimoto (approved):

smorimoto:
From my idea above, this PR looks good, at least to me. Thanks a lot! @shuuji3

ctcpip (approved):

PR #352, opened 2023-10-29
Added translations for proposals.community_translations

Ritish134:
Added German Translation, Japanese Translation, Traditional Chinese 
Fixes #320

PR #398, opened 2024-02-22
✨ don't link to specific agenda and notes

ctcpip:
## What

This PR replaces the links to the upcoming agenda and latest meeting notes  to links to the agendas list and root of the meeting notes tree.

## Why?

- The upcoming agenda is only useful/interesting for 1-2 weeks before plenary.  This means that if updated ASAP, then it will be practically useless for several weeks, and if updated late, then it will just show the old agenda until it gets updated.
- We are assuming people coming to the website will be most interested in the most recent meeting notes, which is overly presumptive.
- We lack automation for updating these links.  While I was originally coming here to update the links to the upcoming agenda and latest meeting notes, it seemed like this could be a better change overall.

## Issues

We need updated translations, but that needn't block merge.

### DIFF
--- a/zh-Hans/site.json
+++ b/zh-Hans/site.json
@@ -4,8 +4,8 @@
   "navigation": {
     "TC39": {
       "title": "TC39",
-      "agenda": "下一次会议的议程",
-      "notes": "历史会议记录",
+      "agenda": "Meeting Agendas",
+      "notes": "Meeting Notes",
       "sub-menu": "展开 “TC39” 的子列表"
     },
     "specs": {
--- a/_data/site.json
+++ b/_data/site.json
@@ -4,8 +4,8 @@
   "navigation": {
     "TC39": {
       "title": "TC39",
-      "agenda": "Upcoming Agenda",
-      "notes": "Recent Meeting Notes",
+      "agenda": "Meeting Agendas",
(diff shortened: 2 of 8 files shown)


zh-Hans/site.json
```diff
@@ -4,8 +4,8 @@
   "navigation": {
     "TC39": {
       "title": "TC39",
-      "agenda": "下一次会议的议程",
-      "notes": "历史会议记录",
+      "agenda": "Meeting Agendas",
+      "notes": "Meeting Notes",
```
ljharb, line 8: i think the translations need to be in "not english"

ctcpip: yes.  see `Issues` above

ljharb: ah. any reason not to just run them through google translate?

ctcpip: > ah. any reason not to just run them through google translate?

https://github.com/tc39/tc39.github.io/pull/352#issuecomment-1793435099

robpalme (dismissed):
I think we should not block on translations.

smorimoto:
@ctcpip Sorry, my PR and this PR seem to have conflicted 😐

ctcpip:
no worries, I'll take care of it

PR #418, opened 2024-04-09
Biome

### LINKED ISSUE #407: Migrate to Biome
https://biomejs.dev

smorimoto:
There doesn't seem to be much point in this project to bridge the gap between ESLint and Prettier and have multiple configuration files. In addition to the large number of rules already ported to Biome, it is also well-compatible with Prettier, which minimises migration noise.


ryzokuken (dismissed):
Overall LGTM, perhaps not so productive to discuss whether it's nicer for us to use JS-based tools vs Rust-based tools but ultimately the smaller configuration is a definite nice-to-have.

smorimoto:
In fact, the debate about whether it's JS-based or Rust-based is pretty pointless.

ctcpip:
double quotes?  oh no  😳

Issue #406, opened 2024-03-14
Add translations for `site.navigation.TC39.agenda` / `site.navigation.TC39.notes`


smorimoto:
### Translation status of each language

Anyone interested in working on this, raise your hand on this issue. (To prevent duplication of effort among multiple contributors!)

| Language             | Status | Link                                                                                                         |
| -------------------- | ------ | ------------------------------------------------------------------------------------------------------------ |
| German Translation   | 🚧     | https://github.com/tc39/tc39.github.io/blob/be8c92e9363633ae018d814c7da88fea5e728e3c/de/site.json#L7-L8      |
| French Translation   | 🚧     | https://github.com/tc39/tc39.github.io/blob/be8c92e9363633ae018d814c7da88fea5e728e3c/fr/site.json#L7-L8      |
| Japanese Translation | ✅     | https://github.com/tc39/tc39.github.io/blob/be8c92e9363633ae018d814c7da88fea5e728e3c/ja/site.json#L7-L8      |
| Russian Translation  | ✅     | https://github.com/tc39/tc39.github.io/blob/be8c92e9363633ae018d814c7da88fea5e728e3c/ru/site.json#L7-L8      |
| Simplified Chinese   | ✅     | https://github.com/tc39/tc39.github.io/blob/be8c92e9363633ae018d814c7da88fea5e728e3c/zh-Hans/site.json#L7-L8 |

Jack-Works:
It looks like there are only 2 lines that need to be translated?

smorimoto:
@Jack-Works Yes

smorimoto:
Well, there's still one more: #320

Jack-Works:
btw zh-Han**s** is Han Simplified, Traditional Chinese should be zh-Han**t**. I think we can also add zh-Hant if there is a need and it's simple to do so.

Jack-Works:
Opened #417 for zh-hans

0xDevNinja:
Hi — taking this one.

After auditing every locale on `main`, French was the only remaining language still showing the English placeholders (`Meeting Agendas` / `Meeting Notes`); the rest are already translated:

| Locale  | `agenda`              | `notes`                |
| ------- | --------------------- | ---------------------- |
| de      | Tagesordnung          | Mitschriften           |
| ja      | 議題                  | 議事録                 |
| ru      | Повестки встреч       | Протоколы встреч       |
| zh-Hans | 会议日程              | 会议记录               |
[...]

Issue #320, opened 2023-04-16
Add translations for `proposals.community_translations`


smorimoto:
### Translation status of each language

At the time this issue was created, work has already been completed in French and Russian, but not yet in the other languages. Anyone interested in working on this, raise your hand on this issue. (To prevent duplication of effort among multiple contributors!)

| Language             | Status | Link                                                                                                            |
| -------------------- | ------ | --------------------------------------------------------------------------------------------------------------- |
| German Translation   | ✅     | https://github.com/tc39/tc39.github.io/blob/be8c92e9363633ae018d814c7da88fea5e728e3c/de/proposals.json#L29      |
| French Translation   | ✅     | https://github.com/tc39/tc39.github.io/blob/be8c92e9363633ae018d814c7da88fea5e728e3c/fr/proposals.json#L29      |
| Japanese Translation | ✅     | https://github.com/tc39/tc39.github.io/blob/be8c92e9363633ae018d814c7da88fea5e728e3c/ja/proposals.json#L29      |
| Russian Translation  | ✅     | https://github.com/tc39/tc39.github.io/blob/be8c92e9363633ae018d814c7da88fea5e728e3c/ru/proposals.json#L29      |
| Simplified Chinese   | ✅     | https://github.com/tc39/tc39.github.io/blob/be8c92e9363633ae018d814c7da88fea5e728e3c/zh-Hans/proposals.json#L29 |

Ritish134:
Hi there would like to work on this issue can you assign it to me

0xDevNinja:
Hi maintainers — quick audit on `main`: every locale's `proposals.json` already has a translated `community_translations` value (no English placeholder remains). It looks like this issue is now stale.

| Locale | `community_translations` excerpt |
| --- | --- |
| `de` | "Diese Seite wird durch die Allgemeinheit übersetzt…" |
| `fr` | "Cette page est traduite par la communauté…" |
| `ja` | "このページはコミュニティによって翻訳されています…" |
| `ru` | "Эта страница переведена сообществом…" |
| `zh-Hans` | "这个页面的翻译是由社区贡献的…" |
| `uk` | "Ця сторінка перекладена спільнотою…" |

A few of the historical commits where translations landed, for reference:

- `e812c7a` — French translation
- `f942934` — Russian translation additions (Sergey Rubanov)
- `ce9b9f6` — Add more translations for `ja/proposals.json`
- `4b9d3aa` — Updated German translation
[...]

PR #435, opened 2024-05-29
Updated german translation

NNowakowski:
Updated german translations for #320 #406 and rephrased footer.connect

### DIFF
--- a/de/site.json
+++ b/de/site.json
@@ -4,8 +4,8 @@
   "navigation": {
     "TC39": {
       "title": "TC39",
-      "agenda": "Meeting Agendas",
-      "notes": "Meeting Notes",
+      "agenda": "Meeting-Agenden",
+      "notes": "Meeting-Notizen",
       "sub-menu": "TC39 Untermenü umschalten"
     },
     "specs": {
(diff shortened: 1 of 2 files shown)


de/site.json
```diff
@@ -4,8 +4,8 @@
   "navigation": {
     "TC39": {
       "title": "TC39",
-      "agenda": "Meeting Agendas",
-      "notes": "Meeting Notes",
+      "agenda": "Meeting-Agenden",
+      "notes": "Meeting-Notizen",
```
TomKopp: I would translate the agendas as "Tagesordnungen" because a "meeting agenda" is a "Tagesordnung".

I can understand the notion to keep the reference to the meetings and be consistent with "Meeting-Notizen" - and "Meeting" is perfectly fine in German.

I think "Meeting-Notizen" could also be "Mitschriften", **if** we go away from "Meeting-*".

NNowakowski (commented):
Updated meeting-* translations

de/site.json
NNowakowski: Good suggestions. I've struggled quite a bit to come up with a good translation for agenda, as that's simply how we call it at my workplace aswell.

TomKopp: We do too. :)

TomKopp (approved):

smorimoto:
Thanks a lot!

PR #456, opened 2024-09-10
Introduce the Ukrainian language

JSMonk:


translation_uk.md
smorimoto, line 1: I'm now thinking that language-specific translation guides are unnecessary. The translator should be able to read English and copy-pasting in each language seems to be useless.

_data/languages.json
```diff
@@ -19,6 +19,10 @@
     "title": "Русский",
     "id": "ru"
   },
+  {
+    "title": "Українська",
+    "id": "uk"
```
smorimoto, line 24: I'm not sure, but isn't it `uk-UA`?

```suggestion
    "id": "uk-UA"
```

chicoxyzzy: ISO 639-1 code is `uk` and country code `UA` shouldn't be necessary

chicoxyzzy (approved):
I am not a native speaker of Ukrainian, but based on my knowledge of the language, I can confirm that the translation seems correct.

smorimoto (approved):

PR #478, opened 2024-11-08
ESM

smorimoto:


de/de.11tydata.js
```diff
@@ -1,4 +1,8 @@
-module.exports = {
+import { createRequire } from "node:module";
+
+const require = createRequire(import.meta.url);
```
smorimoto, line 3: Ref: https://github.com/11ty/eleventy-dependency-tree-esm/pull/3

smorimoto, line 3: https://github.com/11ty/eleventy-dependency-tree-esm/pull/3

package.json
```diff
@@ -6,6 +6,7 @@
     "type": "git",
     "url": "https://github.com/tc39/tc39.github.io.git"
   },
+  "type": "module",
```
ljharb, line 9: it would have been better to make the files `.mjs` instead of using type module, ftr

smorimoto: Hmm? Why?

ljharb: All type module does is change the parse goal of `.js`; it's not required for ESM, and using a `.mjs` extension may even be faster because the looking up the `type` field takes time.

